PR #9, opened 2023-10-11
Add mixin support

Minecraftschurli:


common/src/main/java/net/neoforged/gradle/common/runs/run/RunImpl.java
```diff
@@ -116,14 +116,12 @@ public Set<TaskProvider<? extends Task>> getTaskDependencies() {
...
     @Override
-    @NotNull
     public final void configure() {
         configure(getName());
     }
 
     @Override
-    @NotNull
-    public final void configure(final String name) {
+    public final void configure(final @NotNull String name) {
```
Technici4n, line 124: `@NotNull` is a bad practice, it should be assumed.

Minecraftschurli: Since the super method has the parameter annotated with `@NotNull` intellij yells at me if I don't annotate it.

Technici4n: How about removing it from the superclass? :P

common/src/main/java/net/neoforged/gradle/common/runtime/definition/CommonRuntimeDefinition.java
```diff
@@ -79,6 +79,9 @@ protected CommonRuntimeDefinition(
     public final TaskProvider<? extends WithOutput> getTask(String name) {
         final String taskName = CommonRuntimeUtils.buildTaskName(this, name);
         if (!taskOutputs.containsKey(taskName)) {
+            if (this instanceof IDelegatingRuntimeDefinition) {
```
marchermans: This is up to the implementing delegator, and should not be here.

common/src/main/java/net/neoforged/gradle/common/runtime/definition/CommonRuntimeDefinition.java
```diff
@@ -160,6 +163,8 @@ public void configureAssociatedTask(@NotNull TaskProvider<? extends Runtime> run
     public abstract TaskProvider<ExtractNatives> getNatives();
 
     public void configureRun(RunImpl run) {
+        run.getExtensions().getExtraProperties().set("runtimeDefinition", this);
```
marchermans: Set this as a full blown extension object not as an extra prop

common/src/main/java/net/neoforged/gradle/common/runtime/extensions/CommonRuntimeExtension.java
```diff
@@ -105,12 +105,8 @@ public final D maybeCreate(final Action<B> configurator) {
...
         final S spec = createSpec(configurator);
 
-        if (GradleInternalUtils.getExtensions(project.getExtensions())
-                .stream()
-                .filter(CommonRuntimeExtension.class::isInstance)
-                .map(extension -> (CommonRuntimeExtension<?,?,?>) extension)
-                .anyMatch(ext -> ext.runtimes.containsKey(spec.getIdentifier())))
-            throw new IllegalArgumentException(String.format("Runtime with identifier '%s' already exists", spec.getIdentifier()));
+        if (project.getExtensions().getByType(RuntimesExtension.class).findDefinitionByNameOrIdentifier(spec.getIdentifier()) != null)
+            throw new IllegalArgumentException(String.format("Runtime with name '%s' already exists", spec.getName()));
```
marchermans: Name duplicates are valid. Identifier ones not.

dsl/mixin/src/main/groovy/net/neoforged/gradle/dsl/mixin/extension/Mixin.groovy
```diff
@@ -0,0 +1,14 @@
...
+import net.minecraftforge.gdi.BaseDSLElement
+import net.minecraftforge.gdi.annotations.DSLProperty
+import org.gradle.api.provider.SetProperty
+
+@CompileStatic
+interface Mixin extends BaseDSLElement<Mixin> {
+    String EXTENSION_NAME = "mixin";
+
+    @DSLProperty
+    SetProperty<String> getConfigs();
```
marchermans, line 13: Should this not be a set of files?

marchermans (changes requested):

dsl/mixin/src/main/groovy/net/neoforged/gradle/dsl/mixin/extension/Mixin.groovy
Minecraftschurli: it actually can't be because it needs to be relative to the resources root

PR #21, opened 2023-10-31
Switch back to using the renamed jar as the base for joined binpatch generation

jpenilla:
At some point between 1.20.1 and 1.20.2, the joined binpatches started being based on a recompiled jar, which is obviously not very useful (either for us when reimplementing binpatch only mode to userdev, or for third party consumers)

The only other use of the JOINED cleanProvider is by stripBinaryPatchedClasses; and it's use case shouldn't care whether we use a recompiled jar.


marchermans:
Actually it does matter.

The renamed jar does not contain forge, as such the stripping does not work any more.

Stripping has to be done from the recompiled jar.

But yeah patch generation has to be done from the renamed jar.

marchermans:
This looks good after testing, so merging. If this breaks I will need to revert it.

marchermans (approved):

PR #28, opened 2023-11-02
Add installer logo property to installer task

sciwhiz12:
This PR adds an installer logo property to the `CreateLegacyInstaller` task, for changing the logo shown on the installer window. By default, the installer logo is taken from the root project's `src/main/resources/neoforged_logo.png` file, which is the logo file in the NeoForged repository.

Future work should be done to create our own version of the installer base with our logo and other enhancements and fixes. See [neoforged/LegacyInstaller](https://github.com/neoforged/LegacyInstaller), which is currently unmodified and unpublished.

Tested by assembling the installer in the NeoForge repository (via the `assemble` task, and looking in `projects/neoforge/build/libs`) and running it.

See https://github.com/neoforged/NeoForge/pull/210#discussion_r1377235593 for why this PR was made here in NeoGradle, and not done on the side of NeoForge (in its buildscript).

### DIFF
--- a/platform/src/main/java/net/neoforged/gradle/platform/tasks/CreateLegacyInstaller.java
+++ b/platform/src/main/java/net/neoforged/gradle/platform/tasks/CreateLegacyInstaller.java
@@ -35,6 +35,10 @@ public abstract class CreateLegacyInstaller extends Zip implements WithOutput, W
         from(getLauncherJson());
         from(getInstallerJson());
         from(getUrlIcon());
+        exclude("big_logo.png");
+        from(getInstallerLogo(), spec -> {
+            spec.rename(original -> "big_logo.png");
+        });
         
         from(getUnixServerArgs(), spec -> {
             spec.rename(name -> "data/unix_args.txt");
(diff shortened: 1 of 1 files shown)


Technici4n (changes requested):
Couldn't you change the `profile.getLogo().set("/big_logo.png")` call in DynamicProjectExtension instead?

sciwhiz12:
I'm not sure what you mean...? Even if we change the path to the logo in the profile JSON, we'd still need to copy over the logo file from the (root) project resources and remove/replace the existing one in the installer base (which is currently the Minecraft Forge logo).

Technici4n:
The way I understand it, LegacyInstaller just takes the value of `profile.getLogo()` as the logo to use (see https://github.com/neoforged/LegacyInstaller/blob/389603193163d1d9416b0b7e7affdd5c807235e6/src/main/java/net/minecraftforge/installer/InstallerPanel.java#L166). So we only need to change that.

sciwhiz12:
We would still need to copy over the logo to the installer JAR, which is what my PR is doing. The name or location of the logo is irrelevant, as you've shown the installer can be pointed to wherever the logo is within the JAR -- my PR makes it so the (NeoForge) logo _is_ in the JAR, rather than the Forge logo.

Technici4n (approved):

marchermans (approved):

PR #30, opened 2023-11-04
VsCode debug support

Nightenom:
Adds .vscode/launch.json writer, rest of VSCode support is more or less same as Eclipse (hopefully)
~Adds processResources Copy task for both VSCode and Eclipse~ was merged as separate PR

Things to solve before merging:
- (DONE) launch.json library artifact - if neoforged wants to overtake repo ownership I'm fine with that, https://github.com/Nightenom/VscLaunchJavaWriter


common/src/main/java/net/neoforged/gradle/common/runs/ide/IdeRunIntegrationManager.java
```diff
@@ -190,11 +190,9 @@ public void vscode(Project project, EclipseModel eclipse)
                 final BatchedLaunchWriter launchWriter = new BatchedLaunchWriter(WritingMode.MODIFY_CURRENT);
                 project.getExtensions().configure(RunsConstants.Extensions.RUNS, (Action<NamedDomainObjectContainer<Run>>) runs -> runs.getAsMap().forEach((name, run) -> {
                     final String runName = StringUtils.capitalize(project.getName() + " - " + StringUtils.capitalize(name.replace(" ", "-")));
-                    
                     final RunImpl runImpl = (RunImpl) run;
 
-                    // TODO: what is this and how to hook it in? tasks.json?
-                    final TaskProvider<?> ideBeforeRunTask = createIdeBeforeRunTask(project, name, run, runImpl);
+                    eclipse.autoBuildTasks(createIdeBeforeRunTask(project, name, run, runImpl));
```
marchermans: This is not an autoBuildTask.

This is a task that needs to be ran **before** the actually run task. Please configure it as such.

marchermans (changes requested):

common/src/main/java/net/neoforged/gradle/common/runs/ide/IdeRunIntegrationManager.java
Nightenom: well, right now there is no way to run gradle tasks from IDE's gradle [(see this issue)](https://github.com/microsoft/vscode-java-debug/issues/1106) - the issue apparently isn't high priority nor there is any progress info in repos elsewhere (eclipse/redhat repos, etc)

 I can easily run gradle task in command line via tasks.json though (which would require this tasks to have name - which should be doable? afaik the relevant ide code is also set up during invocation from command line).
 
I went with `autoBuildTasks` because it should provide logically same state (as long as the added task is called after the eclipse build task (which happens after any source file save-to-disc action)), but indeed doing so by calling the task much more times than required

HeatherComputer: > well, right now there is no way to run gradle tasks from IDE's gradle

Huh? Wasn't this exact issue solved in neo 1.20.1 using a prelaunchtask?

```
      "preLaunchTask": "gradle: copyVSCodeResources"
```

I have this as part of my runClient etc runs and it works perfectly fine, this has been required for me both in neo and fabric for recent versions.


(as a little edit, I have no issue debugging the runClient and runServer gradle tasks, which is nice as those haven't worked in the past haha)

Nightenom: Yeah, this was possible in ng6, now there is a task that is created on demand and not available from CLI unless running `gradlew runXYZ`

Btw I assume your prelaunch task is cli gradle call to custom processResources made by you, if you're using microsoft gradle extension than I might add hook into it, but this extension is not part of the java pack and is not needed, ideally I would stick to having only redhat java and vscode-java-debug

I could potentially use attach debug on runXYZ if there won't be any other way, but I leave this as last option cuz it has much longer loading time, although calling the new task from cli (which is what I'm trying rn) may suffer from roughly same delay as attach to runXYZ

marchermans: > Yeah, this was possible in ng6, now there is a task that is created on demand and not available from CLI unless running `gradlew runXYZ`
> 
> Btw I assume your prelaunch task is cli gradle call to custom processResources made by you, if you're using microsoft gradle extension than I might add hook into it, but this extension is not part of the java pack and is not needed, ideally I would stick to having only redhat java and vscode-java-debug
> 
> I could potentially use attach debug on runXYZ if there won't be any other way, but I leave this as last option cuz it has much longer loading time, although calling the new task from cli (which is what I'm trying rn) may suffer from roughly same delay as attach to runXYZ

The runIdexxxx task is available to vscode.... Assuming it sets the right flags when it does the run. If it does not set the flag, then there is no way for us to support this ide properly within the specification of Gradle Toolkit, which we are bound to.

So please verify why you think the ide does not have the runIdeXxx task available.

HeatherComputer: > Btw I assume your prelaunch task is cli gradle call to custom processResources made by you, if you're using microsoft gradle extension than I might add hook into it, but this extension is not part of the java pack and is not needed, ideally I would stick to having only redhat java and vscode-java-debug

Huh, yeah that's part of the microsoft gradle extension, I was always under the impression that it was part of the java pack but I guess not... and yeah the `copyVSCodeResources` task there is defined in my build.gradle but isn't really relevant here.

Nightenom:
Okay, probably my last try on `ideBeforeRunXYZ` task (running out of ideas on what to do next...), also added Copy tasks for processResources for both Eclipse and VSCode

common/build.gradle
```diff
@@ -21,6 +21,7 @@ dependencies {
     api "codechicken:DiffPatch:${project.diffpatch_version}"
     api "commons-codec:commons-codec:${project.commons_codec_version}"
     api "net.neoforged:EclipseLaunchConfigs:${project.eclipse_launch_configs_version}"
+    api "cz.nightenom:vsc-launch-writer:${project.vscode_launch_configs_version}"
```
marchermans: We will need to adopt this plugin into the neoforge org and publish it. Would that be possible?

marchermans (changes requested):

common/build.gradle
Nightenom: As said in PR desc: yes, DM on discord and i can give you ownership

HeatherComputer:
Is there any way I could use this at present? I'd love to be able to debug 1.20.4, which I can't currently do without IDEA (or manually crafting the runs I guess?)

Matyrobbrt:
There is PR publishing, but the author (@Nightenom) will have to merge latest to trigger it.

Nightenom:
It's stale on @marchermans 's side

marchermans:
Sorry wrong button

PR #38, opened 2023-11-08
Re-enable functional tests

shartte:


marchermans:
That is actually on my plan, but I wanted to rewrite them to the new scheme in training wheels. 
There is now a builder based approach to this, which should allow making these tests a lot easier.

Do you just want to focus on enabling them, or do you want to also work on porting them to that scheme?

shartte:
Oh no trouble, I can port to that new scheme, I just wasn't sure if it was the future or not, heh

marchermans:
> Oh no trouble, I can port to that new scheme, I just wasn't sure if it was the future or not, heh

They sure as hell are. Hey support more scenarios and have a much simpler handling.

marchermans:
This has been implemented.

PR #42, opened 2023-11-10
Add customization for max-memory of NeoForm steps

shartte:
…ps via Gradle project properties. Initial uses: Override maximum memory and the tool version used.


shartte:
Will try an approach specifically targetting the decompiler

PR #46, opened 2023-11-11
Transform decompile step into a task that is customizable via specific settings

shartte:
TODO 

- [x] Convention the properties to Gradle properties so they're easy to set via gradle.properties (e.g. `neogradle.subsystems.decompiler.maxMemory=4g`)

Example:

```
subsystems {
  decompiler {
    maxMemory = "2g"
    maxThreads = 4
    logLevel = DecompilerLogLevel.TRACE
    jvmArgs = ["-verbose:gc']
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,21 @@ NeoGradle
 
 Minecraft mod development framework used by NeoForge and FML for the Gradle build system.
 
-For a quick start, see how the [NeoForge Mod Development Kit](https://github.com/neoforged/MDK) uses NeoGradle, or see our official [Documentation](https://docs.neoforged.net/neogradle/docs/).
+For a quick start, see how the [NeoForge Mod Development Kit](https://github.com/neoforged/MDK) uses NeoGradle, or see
+our official [Documentation](https://docs.neoforged.net/neogradle/docs/).
 
 To see the latest available version of NeoGradle, visit the [NeoForged project page](https://projects.neoforged.net/neoforged/neogradle).
+
+## Override Decompiler Settings
+
+The settings used by the decompiler when preparing Minecraft dependencies can be overridden
+using [Gradle properties](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties).
+This can be useful to trade slower build-times for being able to run NeoGradle on lower-end machines.
+
+| Property                                     | Description                                                                                                                |
+|----------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
+| `neogradle.subsystems.decompiler.maxMemory`  | How much heap memory is given to the decompiler. Can be specified either in gigabyte (`4g`) or megabyte (`4096m`).         |
+| `neogradle.subsystems.decompiler.maxThreads` | By default the decompiler uses all available CPU cores. This setting can be used to limit it to a given number of threads. |
+| `neogradle.subsystems.decompiler.logLevel`   | Can be used to override the [decompiler loglevel](https://vineflower.org/usage/#cmdoption-log).                            |
+| `neogradle.subsystems.decompiler.jvmArgs`    | Pass arbitrary JVM arguments to the decompiler. I.e. `-XX:+HeapDumpOnOutOfMemoryError`                                     |
+
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/subsystems/SubsystemsExtension.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/subsystems/SubsystemsExtension.java
@@ -1,0 +1,21 @@
+package net.neoforged.gradle.common.extensions.subsystems;
+
+import net.minecraftforge.gdi.ConfigurableDSLElement;
+import net.neoforged.gradle.dsl.common.extensions.subsystems.Subsystems;
+import org.gradle.api.Project;
+
+import javax.inject.Inject;
+
+public abstract class SubsystemsExtension implements ConfigurableDSLElement<Subsystems>, Subsystems {
+    private final Project project;
+
+    @Inject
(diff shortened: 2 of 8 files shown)


shartte:
@marchermans Added the property binding and at least added some docs to README.md about them, otherwise we'll surely forget to ever document them 😓

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/extensions/subsystems/Decompiler.groovy
```diff
@@ -0,0 +1,52 @@
...
+    Property<Integer> getMaxThreads();
+
+    /**
+     * The log-level to use for the decompiler. Supported values: info, debug, warn, error.
+     * Defaults to info.
+     */
+    @Input
+    @Optional
+    @DSLProperty
+    Property<String> getLogLevel();
```
Matyrobbrt: Any reason this is not an enum?

common/src/main/java/net/neoforged/gradle/common/extensions/subsystems/SubsystemsExtension.java
```diff
@@ -0,0 +1,29 @@
...
+
+public abstract class SubsystemsExtension implements ConfigurableDSLElement<Subsystems>, Subsystems {
+    private final Project project;
+
+    @Inject
+    public SubsystemsExtension(Project project) {
+        this.project = project;
+
+        ProviderFactory providers = project.getProviders();
+        getDecompiler().getMaxMemory().convention(providers.gradleProperty("neogradle.subsystems.decompiler.maxMemory"));
```
Matyrobbrt: `neogradle.subsystems` should be a constant somewhere

Matyrobbrt (changes requested):

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/extensions/subsystems/Decompiler.groovy
shartte: Yes, the reason we discussed on Discord: We pass this through to Vineflower, which has its own idea of log-levels. If we introduce an enum here, we would need to introduce a potentially Vineflower-version-dependent mapping from whatever our enum is to the allowed Vineflower loglevels.

I don't think that's worth the effort for such an advanced feature.

common/src/main/java/net/neoforged/gradle/common/extensions/subsystems/SubsystemsExtension.java
shartte: That would mean you cannot grep the source code anymore to find where neogradle.subsystems.decompiler.somethingsomething is defined :-D
But I can do that if it's desired.

marchermans: Create a utility class to get the properties for neogradle. Give it a method for the subsystem properties and that should deal with it.

shartte: @Matyrobbrt Introduced property prefix constant + enum for loglevel. Can you re-review? :)

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/extensions/subsystems/Decompiler.groovy
```diff
@@ -0,0 +1,52 @@
...
+     * Allows the maximum number of threads used by the decompiler to be constrained. By default, it will
+     * use all available threads.
+     */
+    @Input
+    @Optional
+    @DSLProperty
+    Property<Integer> getMaxThreads();
+
+    /**
+     * The log-level to use for the decompiler. Supported values: info, debug, warn, error.
```
Matyrobbrt:
```suggestion
     * The log-level to use for the decompiler. Supported values: trace, info, warn, error.
```

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/extensions/subsystems/Subsystems.groovy
```diff
@@ -0,0 +1,21 @@
...
+import org.gradle.api.tasks.Nested
+
+/**
+ * Allows configuration of various NeoGradle subsystems.
+ */
+@CompileStatic
+interface Subsystems extends BaseDSLElement<Subsystems> {
+
+    /**
+     * @return Settings for the decompiler subsystem.
```
Matyrobbrt:
```suggestion
     * @return settings for the decompiler subsystem
```

Matyrobbrt (changes requested):

README.md
```diff
@@ -3,6 +3,22 @@ NeoGradle
 
 Minecraft mod development framework used by NeoForge and FML for the Gradle build system.
 
-For a quick start, see how the [NeoForge Mod Development Kit](https://github.com/neoforged/MDK) uses NeoGradle, or see our official [Documentation](https://docs.neoforged.net/neogradle/docs/).
+For a quick start, see how the [NeoForge Mod Development Kit](https://github.com/neoforged/MDK) uses NeoGradle, or see
+our official [Documentation](https://docs.neoforged.net/neogradle/docs/).
+
+To see the latest available version of NeoGradle, visit
+the [NeoForged Maven](https://maven.neoforged.net/#/releases/net/neoforged/gradle/userdev).
```
Matyrobbrt: This should link to `https://projects.neoforged.net/neoforged/neogradle` as it provides copy buttons.

Matyrobbrt (approved):

Matyrobbrt (approved):

Issue #50, opened 2023-11-15
Multiple `ListLibraries` can run in parallel, corrupting each other


SquidDev:
I've a project which applies the userdev plugin in multiple sub-projects. However, attempting to build this project causes errors due to trying to use corrupted library jars.

```
> Task :forge:neoFormRename FAILED
Exception in thread "main" java.lang.RuntimeException: Could not add library: /home/squid/.gradle/caches/minecraft/libraries/org/apache/logging/log4j/log4j-core/2.19.0/log4j-core-2.19.0.jar
        at net.minecraftforge.fart.internal.ClassProviderBuilderImpl.addLibrary(ClassProviderBuilderImpl.java:56)
        at net.minecraftforge.fart.internal.RenamerImpl.lambda$setup$0(RenamerImpl.java:64)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
        at net.minecraftforge.fart.internal.RenamerImpl.setup(RenamerImpl.java:64)
        at net.minecraftforge.fart.internal.RenamerImpl.run(RenamerImpl.java:72)
        at net.minecraftforge.fart.Main.main(Main.java:159)
Caused by: java.util.zip.ZipException: zip END header not found
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.findEND(ZipFileSystem.java:1320)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.initCEN(ZipFileSystem.java:1534)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.<init>(ZipFileSystem.java:179)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getZipFileSystem(ZipFileSystemProvider.java:125)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.newFileSystem(ZipFileSystemProvider.java:120)
        at java.base/java.nio.file.FileSystems.newFileSystem(FileSystems.java:528)
        at java.base/java.nio.file.FileSystems.newFileSystem(FileSystems.java:400)
        at net.minecraftforge.fart.internal.ClassProviderBuilderImpl.addLibrary(ClassProviderBuilderImpl.java:38)
        ... 5 more
```

From what I can tell, the `neoFormListLibraries` task from one project runs in parallel with `neoFormRename` in another, causing the latter to read corrupt partially-downloaded jars.

I think there's several things which should be done to avoid this:
[...]

PR #53, opened 2023-11-19
Check hashes in ListLibraries

SquidDev:
We now use `DownloadFileAction` to download files, rather than doing it directly. This ensures we only download files if they are missing (or the cache is invalid), alleviating some of the issues in #50.

This also does a similar exists+hash check when extracting from a bundle.


marchermans (approved):

SquidDev:
Thanks!

PR #57, opened 2023-11-25
Fix spaces in Gradle user home breaking IntelliJ run configurations

### LINKED ISSUE #54: Cannot run client in dev due to space in my Windows file path
Attempting to run Neoforge with a setup very close to the MDK, I get this error:
```
Connected to the target VM, address: '127.0.0.1:63513', transport: 'socket'
Error: Could not find or load main class Laptop\.gradle\caches\modules-2\files-2.1\org.ow2.asm\asm-commons\9.5\19ab5b5800a3910d30d3a3e64fdb00fd0cb42de0\asm-commons-9.5.jar;C:\Users\MSI
Caused by: java.lang.ClassNotFoundException: Laptop\/gradle\caches\modules-2\files-2/1\org/ow2/asm\asm-commons\9/5\19ab5b5800a3910d30d3a3e64fdb00fd0cb42de0\asm-commons-9/5/jar;C:\Users\MSI
Disconnected from the target VM, address: '127.0.0.1:63513', transport: 'socket'
```

My file path for my project is:
`C:\Users\MSI Laptop\Documents\ModWorkspace\CommandStructures-Neoforge`

The space seems to be what is breaking the project. Not sure if I can remove that space without breaking a bunch of my other non-mc stuff since I had that path since like forever.

shartte:
To reproduce this in IntelliJ:
image

Fixes #54

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/runs/ide/IdeRunIntegrationManager.java
+++ b/common/src/main/java/net/neoforged/gradle/common/runs/ide/IdeRunIntegrationManager.java
@@ -35,6 +35,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


marchermans:
This commit does not explain why it is fixed.....

We were already quoting the programme arguments it is crashing over. What do jvm arguments have to do with this?

Matyrobbrt:
The module classpath is a jvm argument. See: https://discord.com/channels/313125603924639766/801175194298744902/1178110006088974376

shartte:
> This commit does not explain why it is fixed.....
> 
> We were already quoting the programme arguments it is crashing over. What do jvm arguments have to do with this?

In Eclipse Run configurations, the JVM arguments are passed as an array and Eclipse seems to take care of escaping individual arguments. For IntelliJ, we have to do that. And as maty says, the module path (which contains paths pointing into GRADLE_USER_HOME)  will be a JVM argument.

PR #62, opened 2023-11-30
Remove the ignore list regex

Matyrobbrt:
The regex attempts to remove version information from the jar, so that it would go from `artifact-name-<version>.jar` to `artifact-name` in the BSL ignore list. But the full name will work as well. See https://github.com/McModLauncher/bootstraplauncher/blob/d1e91742561c9ea37c035c0fec2bd756967a1395/src/main/java/cpw/mods/bootstraplauncher/BootstrapLauncher.java#L70.

And in fact the full name is already being used by fml-core and fml-events (jar names `core-<version>.jar` and `events-<version>.jar`) because `core` is too generic of a prefix to be excluding, and same for `event`.

It was buggy when paired with versions with classifiers, and is simply not needed. No implementations of maven downloaders will rename the file. So just remove the core/events special cases and the regex altogether.


Technici4n (approved):
Tested forgedev, userdev, prod.

PR #65, opened 2023-12-05
Implement a fallback when including non-NeoGradle source sets in a run

shartte:
Implement a fallback when including source sets that do not have NeoGradle applied so they get merged with the sources from the project in which the runs are defined.

This fixes issues with the multi-loader setups (i.e. JEIs) where a Common project is managed by sponges VanillaGradle, while the NeoForge projects are managed by NeoGradle.

In this case, the mod author just wants to have all their projects merged into a single mod at runtime:
- project(":Common").sourceSets.main [VanillaGradle]
- project(":CommonApi").sourceSets.main [VanillaGradle]
- project(":NeoForgeApi").sourceSets.main [NeoGradle]
- project(":NeoForge").sourceSets.main [NeoGradle]

With the changes in this PR, this becomes possible by specifying a mod-id on the NeoGradle and NeoGradleApi projects and  including the source-sets of Common and CommonApi as usual, since these two projects will then inherit the mod id `jei`.

```
minecraft {
	modIdentifier.set("jei")
}
```

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/runs/ide/IdeRunIntegrationManager.java
+++ b/common/src/main/java/net/neoforged/gradle/common/runs/ide/IdeRunIntegrationManager.java
@@ -37,7 +37,7 @@ import java.io.UncheckedIOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


marchermans:
Do I understand this right that this makes any unknown sourceset use the project that owns the runs mod identifier?

shartte:
> Do I understand this right that this makes any unknown sourceset use the project that owns the runs mod identifier?

Yes, exactly

shartte:
Another possibility would be something along those lines...
I don't remember if FG5 or 6 had this, or if I am completely misremembering...

```
runs {
  configureEach {
    mods {
      jei {
        sourceSet sourceSets.main
        sourceSet project(":Common").sourceSets.main
        sourceSet project(":CommonApi").sourceSets.main
        sourceSet project(":NeoForgeApi").sourceSets.main
      }
    }
  }
}
```

marchermans:
> Another possibility would be something along those lines...
> I don't remember if FG5 or 6 had this, or if I am completely misremembering...
> 
> ```
> runs {
>   configureEach {
>     mods {
>       jei {
>         sourceSet sourceSets.main
>         sourceSet project(":Common").sourceSets.main
>         sourceSet project(":CommonApi").sourceSets.main
>         sourceSet project(":NeoForgeApi").sourceSets.main
>       }
>     }
>   }
> }
> ```

You are right, but I explicitly redesigned it, because I really want to allow direct project references during runs, instead of these kinds of hacks. In my eyes, after the platform DSL completes, I want to nuke mod sources from orbit.

shartte:
We need something better / more expressive than this.

PR #75, opened 2023-12-14
Various Task Improvements (Patch, Strip, Recompile)

shartte:
This combines various task improvements aimed at reducing task inputs and thus the work that up-to-date checks and build-cache have to do.

- Use DiffPatches ability to apply patches directly from ZIP-Files and use the NeoForm artifact as the source directly
- Minor cleanup use try-with-resources in Strip
- Don't use doLast in RecompileSources and instead move it in to the Task action
- Don't cache ArtifactFromOutput since it's just a file copy (see Gradles own  `Copy` task for similar reasoning on their part)


marchermans:
@shartte Can you update this?

marchermans (approved):

shartte:
@marchermans Rebased

PR #104, opened 2024-01-15
Don't automatically configure runs when using a `Provider<RunType>`

SquidDev:
This adds the missing `getConfigureFromTypeWithName().set(false)` when configuring runs with a `Provider<RunType>` to make it consistent with the other two `configure(runType)` overrides.

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/runs/run/RunImpl.java
+++ b/common/src/main/java/net/neoforged/gradle/common/runs/run/RunImpl.java
@@ -141,6 +141,7 @@ public abstract class RunImpl implements ConfigurableDSLElement<Run>, Run {
 
     @Override
     public void configure(@NotNull Provider<RunType> typeProvider) {
+        getConfigureFromTypeWithName().set(false); // Don't re-configure
         this.runTypes.add(typeProvider);
     }
 


Technici4n (approved):
Looks like an oversight from https://github.com/neoforged/NeoGradle/commit/902ae2cbc96881bc9ffd93f740ce793c1a966a21, indeed. Previously the method used to delegate to another `configure` method, hence the missing check.

marchermans (approved):

shartte:
@SquidDev Can you give me an idea of how you are using this? This is probably an oversight since the usercases are just not known 😓

SquidDev:
I'm using the Kotlin syntax for runs, where `runTypes.client` is equivalent to `runTypes.named("client")` (returning a provider) rather than getting the run type directly:

```kotlin
runs {
  val gameTestClient by registering {
    configure(runTypes.client) // <- Uses the provider overload
    configureFromTypeWithName = false // FIXME: This shouldn't be required.
  }
}

shartte:
> I'm using the Kotlin syntax for runs, where `runTypes.client` is equivalent to `runTypes.named("client")` (returning a provider) rather than getting the run type directly:
> 
> ```kotlin
> runs {
>   val gameTestClient by registering {
>     configure(runTypes.client) // <- Uses the provider overload
>     configureFromTypeWithName = false // FIXME: This shouldn't be required.
>   }
> }
> ```

Ah, good to know. I think we usually tested with string-based configure() calls

SquidDev:
Thanks!

PR #107, opened 2024-01-24
Add JUnit support

Matyrobbrt:
Add support for configuring JUnit run tasks.


SquidDev:
How does this interact with running tests from IDEA directly (rather than through Gradle)? I'm assuming that it won't work, as FML/BootstrapLauncher expect environment variables and system properties to be present?

Matyrobbrt:
If you run tests with the Gradle runner it will work fine in IDEA, if you run with the IntelliJ runner it won't, that's correct, and that's not really fixable.

SquidDev:
That's useful to know, thanks! Mostly just trying to compare with what I have now (I only need MC's Bootstrap for my existing tests, so can get away with doing less of a thorough job :)).

platform/src/main/java/net/neoforged/gradle/platform/extensions/DynamicProjectExtension.java
```diff
@@ -532,12 +534,29 @@ public void runtime(final String neoFormVersion, Directory patches, Directory re
...
+                type.getArguments().add("forgejunituserdev");
+                type.getArguments().add("--version");
+                type.getArguments().add(project.getVersion().toString());
+                type.getArguments().add("--assetIndex");
+                type.getArguments().add("{asset_index}");
+                type.getArguments().add("--assetsDir");
+                type.getArguments().add("{assets_root}");
+
+                configureUserdevRunType(type, moduleOnlyConfiguration, gameLayerLibraryConfiguration, pluginLayerLibraryConfiguration, userdevCompileOnlyConfiguration, project);
+            });
```
marchermans: It looks like this run type does mark the run as JUnit capable.

junit/src/main/java/net/neoforged/gradle/junit/JUnitProjectPlugin.java
```diff
@@ -0,0 +1,157 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(JUnitProjectPlugin.class);
+
+    @Override
+    public void apply(Project project) {
+        project.getTasks().named("test").configure(task -> task.getExtensions().create(TestTaskExtension.NAME, TestTaskExtension.class, project));
+        project.afterEvaluate(this::applyAfterEvaluate);
+    }
+
+    private void applyAfterEvaluate(Project project) {
+        Test testTask = (Test) project.getTasks().getByName("test");
```
marchermans: This should probably be ran for all test tasks.
Or some kind of extension should exist within the project to control what test tasks it should target.

junit/src/main/java/net/neoforged/gradle/junit/JUnitProjectPlugin.java
```diff
@@ -0,0 +1,157 @@
...
+        DependencyFactory dependencyFactory = project.getDependencyFactory();
+        if (runtimeDefinition instanceof UserDevRuntimeDefinition) {
+            UserDevRuntimeDefinition userdevRuntime = (UserDevRuntimeDefinition) runtimeDefinition;
+            List<String> testDependencies = userdevRuntime.getUserdevConfiguration().getAdditionalTestDependencyArtifactCoordinates().get();
+            for (String testDependency : testDependencies) {
+                testRuntimeOnly.getDependencies().add(dependencyFactory.create(testDependency));
+            }
+        }
+
+        RunImpl junitRun = getOrCreateRun("junitTest", project, runtimeDefinition, runType);
```
marchermans, line 98: No if the user does not create one, then you should abort processing.
The userdev runtime will create a run named after the runtype, so creating another run here is redundant and wrong anyway.

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/runs/type/RunType.groovy
```diff
@@ -103,6 +104,15 @@ abstract class RunType implements ConfigurableDSLElement<RunType>, NamedDSLEleme
     @Input
     abstract Property<Boolean> getIsServer();
 
+    /**
+     * Indicates if this run is a JUnit run.
+     *
+     * @return {@code true} if this run is a JUnit run; otherwise, {@code false}.
+     */
+    @DSLProperty(propertyName = 'junit')
```
marchermans, line 112: Remove the property name. The others do not have it either.

common/src/main/java/net/neoforged/gradle/common/util/run/RunsUtil.java
```diff
@@ -37,21 +42,32 @@ public static String createTaskName(final Run run) {
...
+
+        project.afterEvaluate(evaluatedProject -> {
+            run.getEnvironmentVariables().put("MOD_CLASSES", buildGradleModClasses(run.getModSources()));
+
+            runTask.configure(task -> {
+                addRunSourcesDependenciesToTask(task, run);
+
+                run.getTaskDependencies().forEach(task::dependsOn);
+            });
+        });
```
marchermans: This change is not needed as far as I can tell. If you have a reason for it, please document it accordingly.

marchermans (changes requested):

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/runs/type/RunType.groovy
Matyrobbrt: It is needed, otherwise the generated name is `jUnit`.

marchermans: Hmm okey
I forgot indeed that it does that
Allright thank you.

junit/src/main/java/net/neoforged/gradle/junit/JUnitProjectPlugin.java
shartte: The user is explicitly applying a special Neogradle junit plugin to get junit tests to work. That should get that set up without surprises and as seamlessly as possible, in my opinion.

Also: I do not think the userdev runtime creates runs based on runtypes automatically.

Matyrobbrt: No, it does not create them automatically. And even if it did, that would be a bad idea for the junit run type (and so would the user creating it normally). The run would generate `run*` tasks, `prepare*` tasks, and IDE run configurations which wouldn't work with JUnit.

marchermans: Indeed it does not create the runs automatically. But your runtype should be configured enough that it should only need a: 
```groovy
runs {
   unitTests{} //Or what ever the name of the unit test run type is.
}
```
To get the plugin to work. That is how all other run types are designed.
Please redesign this area.

marchermans (changes requested):

shartte:
@marchermans The IntelliJ & Gradle unit test launchers are not runTypes, hence the comparison falls flat. We do not launch via a generated gradle task, nor do we launch via BootstrapLauncher here.
I don't want to expose these internal shenanigans for users since it's *more* friction to get unit tests to work and not less.

marchermans:
Is superseded by #183

PR #110, opened 2024-01-31
Do not change standard out/error log levels

SquidDev:
`getLogging()` is not thread-safe in parallel builds (https://github.com/gradle/gradle/issues/27897), which can cause standard output to be permanently redirected to `DEBUG` log levels. This is especially unhelp when trying to run Minecraft itself, as you get no output, and so have to tail the log files instead!

Removing these lines fixes this issue, and doesn't appear to have any impact on what information is actually logged during normal usage.


marchermans (approved):

SquidDev:
Thanks!

PR #118, opened 2024-02-19
Prepare for configuration cache

marchermans:


common/src/main/java/net/neoforged/gradle/common/runtime/tasks/ListLibraries.java
```diff
@@ -47,7 +47,11 @@ public ListLibraries() {
             return new File(arguments.get("bundle").get());
         }));
         getOutputFileName().set("libraries.txt");
+        getIsOffline().convention(getProject().getGradle().getStartParameter().isOffline());
```
shartte, line 50: This works? Really? Does setting the offline mode affect the configuration phase at all?

marchermans: Yes this works, I did this already in other places.
The `isOffline` start parameter is configured way before task configuration, and as such its value can be set as the convention.

PR #123, opened 2024-02-27
Fix change in behavior of NeoFormRuntimeSpecification#getNeoFormVersion

lukebemish:
PR #86 caused a change in the behavior of `NeoFormRuntimeSpecification#getNeoFormVersion`, in which it now returns versions in the form `MCVERSION-TIMESTAMP` instead of `TIMESTAMP`. This PR fixes this regression, and updates locations which make use of the new behavior to use `Specification#getVersion` instead, which returns the full MC-version-and-timestamp version string, as well as adding something to the javadoc of the method explaining which form is should return. I have tested this in neodev, and it creates an installer which works, and which can launch the game as a dedicated server successfully.

### DIFF
--- a/neoform/src/main/java/net/neoforged/gradle/neoform/runtime/specification/NeoFormRuntimeSpecification.java
+++ b/neoform/src/main/java/net/neoforged/gradle/neoform/runtime/specification/NeoFormRuntimeSpecification.java
@@ -52,9 +52,9 @@ public class NeoFormRuntimeSpecification extends CommonRuntimeSpecification impl
     public String getMinecraftVersion() {
         return config.getVersion();
     }
-    
+
     public String getNeoFormVersion() {
-        return getVersion();
+        return getVersion().substring(getVersion().lastIndexOf("-") + 1);
     }
 
     public File getNeoFormArchive() {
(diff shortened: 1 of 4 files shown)


marchermans (approved):

neoform/src/main/java/net/neoforged/gradle/neoform/runtime/specification/NeoFormRuntimeSpecification.java
```diff
@@ -52,9 +52,9 @@ public NeoFormConfigConfigurationSpecV2 getConfig() {
     public String getMinecraftVersion() {
         return config.getVersion();
     }
-    
+
     public String getNeoFormVersion() {
-        return getVersion();
+        return getVersion().substring(getVersion().lastIndexOf("-") + 1);
```
shartte, line 57: Doesn't this specifically revert the ability to use a branch-version of neoform successfully or am I misremembering?

lukebemish: Hmm, maybe - how're those formatted? Do we need to use `indexOf` instead here?

PR #126, opened 2024-03-08
Add ability to set primary source set for run

Minecraftschurli:


marchermans:
This is already possible using the idea extension on the run. This is not part of the main run since this property is specific to runs generated for idea

PR #131, opened 2024-03-16
Fix and rethink JarJar task and extension

lukebemish:
The current jarJar task and extension are heavily intertwined with the use of `Dependency`, `Configuration` and related objects as task inputs, making them incompatible with gradle best practices and configuration caching. This PR fixes this; I have made an attempt to keep existing systems or API where they make sense, but substantial areas *have* been rewritten in order to conform with idiomatic gradle approaches. Several improvements include:
- The jarJar system should work with configuration caching now.
- The generated `-all` jar is published as the artifact of `runtimeElements` by default now, meaning that consumers of a library that jarJars other libraries will, at runtime, now by default consume the library containing the jarJared elements - meaning that you can now jarJar a non-mod library and consumers will see it at runtime without having to manually add it to the runs
- Version range detection works with the built-in gradle systems in addition to `jarJar.ranged` and `jarJar.pin`. For example:
```gradle
jarJar("dev.lukebemish.opensesame:opensesame-core:[0.5.1,)") {
    version {
        prefer "0.5.1"
    }
}
```

~~This PR is still a draft for the moment because I need to write some tests for the updated system.~~
Leaving the tests out for now because the test system is broken and cannot actually test whether stuff is config cache compliant. If you'd like me to add other tests for stuff besides that let me know.


userdev/src/main/java/net/neoforged/gradle/userdev/UserDevProjectPlugin.java
```diff
@@ -38,6 +39,13 @@ public void apply(Project project) {
 
     protected void configureJarJarTask(Project project, JarJar jarJarExtension) {
         final Configuration configuration = project.getConfigurations().create(JAR_JAR_DEFAULT_CONFIGURATION_NAME);
+        configuration.setTransitive(false);
+        configuration.getAllDependencies().configureEach(dep ->
+                jarJarExtension.enable()
+        );
+        configuration.attributes(attributes ->
+                attributes.attribute(Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, Usage.JAVA_RUNTIME))
```
marchermans: You can probably also set the contained dependencies attribute, no?

marchermans: Nevermind misread

common/src/main/java/net/neoforged/gradle/common/extensions/JarJarExtension.java
```diff
@@ -35,9 +39,30 @@ public void enable() {
...
+        if (this.enabled == enabled) {
+            return;
+        }
+        this.enabled = enabled;
+        final JarJar task = (JarJar) project.getTasks().findByPath("jarJar");
+        Configuration runtimeElements = project.getConfigurations().maybeCreate(JavaPlugin.RUNTIME_ELEMENTS_CONFIGURATION_NAME);
         if (task != null) {
-            task.setEnabled(enabled);
+            if (enabled) {
+                removedFromPublication.clear();
```
lukebemish: Leaving a note to myself here; I have to actually add the new stuff to this list to make enabling jarJar reversible

common/src/main/java/net/neoforged/gradle/common/dependency/AbstractDependencyManagementObject.java
```diff
@@ -4,47 +4,64 @@
...
+        return new ArtifactIdentifier(identifier.getModuleIdentifier().getGroup(), identifier.getModuleIdentifier().getName(), identifier.getVersion());
     }
 
-    protected static ArtifactIdentifier createArtifactIdentifier(final ModuleDependency dependency) {
+    protected static ArtifactIdentifier createArtifactIdentifier(final Dependency dependency) {
         return new ArtifactIdentifier(dependency.getGroup(), dependency.getName(), dependency.getVersion());
     }
 
+    @SuppressWarnings("unchecked")
     public Spec<? super ArtifactIdentifier> dependency(Object notation) {
```
lukebemish: Gradle is moving in the direction of typesafe dependency creation methods instead of the generic Object one (see, for example, the new DependencyFactory API). Might be worth changing this to match, and/or exposing the dependency factory itself somehow instead (do this via a closure/action mayhaps? I'll think about it)

lukebemish:
~~While working on a similar PR for loom, I've discovered some things I ought to test and possibly modify - namely, how project dependencies behave with this system.~~

This has been fixed, by inspecting first the module, and if it is missing the first declared capability. There is something to be said for doing this in the opposite order, or picking the module if it is a capability otherwise the first capability, but I am unsure, so that might get changed up a bit more

marchermans (approved):

PR #134, opened 2024-03-23
Publish with gradle API attribute

lukebemish:
Adds the `org.gradle.plugin.api-version` attribute to the published elements with a value of the gradle version compiled against; I was unable to find any simpler way of doing this, unfortunately.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,6 @@
+import org.gradle.api.attributes.plugin.GradlePluginApiVersion
+import org.gradle.util.GradleVersion
+
 plugins {
     id 'eclipse'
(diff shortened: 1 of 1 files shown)


lukebemish:
Gradle adds nice error reporting for this attribute in [8.7](https://docs.gradle.org/8.7/release-notes.html#:~:text=Improved%20plugin%20application%20error%20reporting), so users on older versions will get an attribute resolution error instead of the nice on introduced there; however, this is still better than whatever obscure bugs might arise from using it on too old a gradle version, where no proper error reporting is present.

marchermans (approved):

PR #140, opened 2024-04-05
Implement delayed resolve of dev config artifacts.

marchermans:
## Removal of unpacked NeoForm and Userdev artifacts.
This PR removes the unpacked neoform and userdev artifacts in the build directory.
As such it allows clean and build to be executed in the same execution.

## Update to new training wheels
This PR applies an update to training wheels allowing for better testing and better control of how to test.


dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/MinecraftVersionAndUrl.groovy
```diff
@@ -0,0 +1,34 @@
...
+     * The full version name of the minecraft version.
+     *
+     * @return The full version name of the minecraft version.
+     */
+    String getVersion() {
+        return version
+    }
+
+    /**
+     * The url to download the minecraft version from.
```
shartte: I think this should clarify what the URL points to.
The version manifest?

common/src/main/java/net/neoforged/gradle/common/extensions/MinecraftArtifactCacheExtension.java
```diff
@@ -87,44 +89,36 @@ public final Map<CacheFileSelector, File> getCacheFiles() {
...
-        GameArtifact.LAUNCHER_MANIFEST.doWhenRequired(side, () -> result.put(GameArtifact.LAUNCHER_MANIFEST, this.cacheLauncherMetadata()));
-        GameArtifact.VERSION_MANIFEST.doWhenRequired(side, () -> result.put(GameArtifact.VERSION_MANIFEST, this.cacheVersionManifest(finalGameVersion)));
-        GameArtifact.CLIENT_JAR.doWhenRequired(side, () -> result.put(GameArtifact.CLIENT_JAR, this.cacheVersionArtifact(finalGameVersion, DistributionType.CLIENT)));
-        GameArtifact.SERVER_JAR.doWhenRequired(side, () -> result.put(GameArtifact.SERVER_JAR, this.cacheVersionArtifact(finalGameVersion, DistributionType.SERVER)));
-        GameArtifact.CLIENT_MAPPINGS.doWhenRequired(side, () -> result.put(GameArtifact.CLIENT_MAPPINGS, this.cacheVersionMappings(finalGameVersion, DistributionType.CLIENT)));
-        GameArtifact.SERVER_MAPPINGS.doWhenRequired(side, () -> result.put(GameArtifact.SERVER_MAPPINGS, this.cacheVersionMappings(finalGameVersion, DistributionType.SERVER)));
+        GameArtifact.VERSION_MANIFEST.doWhenRequired(side, () -> result.put(GameArtifact.VERSION_MANIFEST, this.cacheVersionManifest(resolvedVersion.getVersion())));
```
shartte: When resolvedVersion now contains `url`, why is this re-resolving? 🤔 If `url` points to the version manifest, we should just pass that here. (Or are we assuming it is cached anyway).

PR #146, opened 2024-04-17
Add system for publishing and consuming ATs

lukebemish:
Currently, a limit of access transformers is that, if you wish to use a vanilla class you've ATed in a mod's API, every dependent must independently apply the AT in question. This PR offers a solution to this, that does not have the worries associated with it of "transitive ATs" or similar concepts, and is no less safe in terms of modders forgetting a `mods.toml` dependency than existing `implementation` dependencies are.

## Goals

- allow mods built as multi-project systems to easily share ATs between subprojects
- allow mods to publish AT files in a consumable fashion
- allow mods to consume specifically published ATs of specific modules

## Non-goals
- automatic detection of ATs from dependencies -- depending on an AT requires an explicit dependency on the ATs of the module, avoiding any sort of automatic detection, mitigating the chances of a modder depending on an AT of an optional dependency that is not present in production.
- Any sort of discovery from dependency jars -- `fg.deobf` sucked and we don't want it back.

To expose an AT file, a mod would do the following:
```gradle
accessTransformers {
    expose project.file("src/main/resources/META-INF/accesstransformers.cfg")
}
```
This publishes the AT file as a new variant of the project with the `accesstransformer` category.

To consume the AT of a module requires an additional dependency on it beyond the normal `implementation` or the like:
```gradle
dependencies {
    implementation "org.example:somemod:0.1.0"
    implementation "org.example:anothermod:0.2.0"
}

accessTransformers {
    consume "org.example:somemod:0.1.0" // pulls in the published AT file of org.example:somemod
    consumeApi "org.example:anothermod:0.2.0" // pulls in the published AT file of org.example:anothermod, and additionally exposes it transitively in accessTransformerElements
}
```
Note that the explicit need to declare a dependency on the access transformers of the module makes this no less safe, in terms of forgetting to declare a `mods.toml` dependency, than a normal `implementation` dependency on a mod and the use of its classes would; a mod cannot "accidentally" consume a foreign AT file.


Technici4n:
I don't like this feature for the simple reason that it is niche and requires advanced Gradle knowledge to use or understand. If we add many such features the code will become unmaintainable.

lukebemish:
What advanced knowledge does this require to use? It was built expressly to require very little gradle knowledge to use.

PR #147, opened 2024-04-18
Implement ATs with JST and enable incremental compilation

Matyrobbrt:
Apply ATs on the source and not on the pre-decomp jar, to allow AT changes to be applied quicker, by not involving decompilation again.

To incrementally compile, the task should only detect changes in files on disk. To achieve this, the sources are unpacked before compilation, in such a way that files aren't overwritten unless they need to be, as timestamps play a role in verifying whether a file change should be ignored.  
This PR also makes the list libraries task deterministic.


neoform/src/main/java/net/neoforged/gradle/neoform/runtime/tasks/UnpackJavaSources.java
```diff
@@ -0,0 +1,90 @@
...
+
+@CacheableTask
+public abstract class UnpackJavaSources extends DefaultRuntime {
+
+    public UnpackJavaSources() {
+        getUnpackingTarget().convention(getOutputDirectory().map(dir -> dir.dir("unpacked")));
+    }
+
+    @TaskAction
+    public void doTask() throws IOException {
```
lukebemish, line 38: This feels like something to just use `FileSystemOperations` and a copy spec for instead of manually traversing stuff

lukebemish (changes requested):

marchermans (approved):

PR #149, opened 2024-04-27
Make jarJar configuration not consumable

lukebemish:
Currently, the `jarJar` configuration is consumable. As it has attributes, this can lead to weirdness when consuming a project that uses jarJar via a project dependency. This fixes that; the configuration has no need to be consumable.

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/DefaultJarJarFeature.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/DefaultJarJarFeature.java
@@ -119,6 +119,8 @@ public class DefaultJarJarFeature implements JarJarFeature {
         configuration.getAllDependencies().configureEach(dep ->
                 this.enable()
         );
+        // jarJar configurations should be resolvable, but ought not to be exposed to consumers;
+        // as it has attributes, it could conflict with normal exposed configurations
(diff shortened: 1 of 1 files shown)


marchermans (approved):

common/src/main/java/net/neoforged/gradle/common/extensions/DefaultJarJarFeature.java
```diff
@@ -119,6 +119,8 @@ public void createTaskAndConfiguration() {
         configuration.getAllDependencies().configureEach(dep ->
                 this.enable()
         );
+        configuration.setCanBeResolved(true);
+        configuration.setCanBeConsumed(false);
```
marchermans, line 125: In second thought, please leave a comment as to why this is needed.

marchermans (changes requested):

common/src/main/java/net/neoforged/gradle/common/extensions/DefaultJarJarFeature.java
lukebemish: Done

shartte (approved):

shartte (approved):

PR #151, opened 2024-04-27
Fix string delimiters for Gradle dependencies in README.md

qtdyn:
:)


Technici4n:
The `"` is required for string interpolation, so no.

dynpatch:
So it can be changed to both double-quotes then?

Technici4n:
Ahhh yes that is fine

dynpatch:
Sorry I wasn't thinking of interpolation when changing this :sweat:

Technici4n:
Apparently Orion already fixed this?

dynpatch:
Yup, case closed

PR #158, opened 2024-04-28
Implement native support for version catalogues

### LINKED ISSUE #141: Support gradle version catalogues for run config dependencies
Currently, NeoGradle does not properly support [gradle version catalogues](https://docs.gradle.org/current/userguide/platforms.html#sub:version-catalog) and this syntax will fail to configure:
```gradle
runs {
    configureEach {
    	// ...
    	dependencies {
    		runtime libs.emi.neoforge
    	}
    }
 }
```

Error:
`Cannot convert the provided notation to an object of type Dependency: map(valueof(DependencyValueSource)).`

This can be worked around by converting the dependency to string notation (via `libs.emi.neoforge.get()`), but implementing proper support would be appreciated.

marchermans:
This adds native support for version catalogue entries when using run dependencies.

The new format as such now supports:

```groovy
runs {
   client {
      dependencies {
         runtime libs.jgrapht
      }
   }
}
```

Closes: #141


dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/ConfigurationUtils.groovy
```diff
@@ -11,13 +12,40 @@ import org.gradle.api.provider.Provider
...
 
+    /**
+     * Creates an attached configuration that can be resolved, but not consumed.
+     *
+     * @param project The project to create the configuration for
+     * @param dependencies The dependencies to add to the configuration
+     * @return The detached configuration
+     */
+    static Configuration temporaryAttachedConfiguration(final Project project, final Dependency... dependencies) {
+        final Configuration configuration = project.getConfigurations().create("ng_do_not_use_this_configuration_${temporaryConfigurationCounter.incrementAndGet()}", new Action<Configuration>() {
```
shartte: Oh come on, Orion... Your previous argument was, that people could not discover these "magic configurations" naturally. Now you provide a nice helper, and make this mess :-P
Just give them sensible names. "run<blah>RuntimeClasspath", for example.

PR #161, opened 2024-04-30
Avoid deleting dummy repository files unless needed

Matyrobbrt:
This practice not only breaks caching, but also causes issues with concurrent invocations of Gradle tasks (i.e. IDE refresh).

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/repository/IvyDummyRepositoryExtension.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/repository/IvyDummyRepositoryExtension.java
@@ -169,13 +169,13 @@ public abstract class IvyDummyRepositoryExtension implements ConfigurableDSLElem
         writeIvyMetadataFile(entry, jarFile, baseDir, metaFile);
 
         if (!Files.isRegularFile(jarFile)) {
-            Files.deleteIfExists(jarFile);
+            FileUtils.delete(jarFile);
             Files.createFile(jarFile);
         }
 
         final Path sourcesFile = entry.asSources().buildArtifactPath(getRepositoryDirectory().get().getAsFile().toPath());
         if (!Files.isRegularFile(sourcesFile)) {
-            Files.deleteIfExists(sourcesFile);
+            FileUtils.delete(sourcesFile);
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/neoforged/gradle/common/extensions/repository/IvyDummyRepositoryExtension.java
```diff
@@ -168,12 +168,16 @@ private void writeDummyDataIfNeeded(
 
         writeIvyMetadataFile(entry, jarFile, baseDir, metaFile);
 
-        Files.deleteIfExists(jarFile);
-        Files.createFile(jarFile);
+        if (!Files.isRegularFile(jarFile)) {
+            Files.deleteIfExists(jarFile);
```
shartte: What is the point of deleteIfExists if it's not a file

common/src/main/java/net/neoforged/gradle/common/extensions/repository/IvyDummyRepositoryExtension.java
```diff
@@ -168,12 +168,16 @@ private void writeDummyDataIfNeeded(
...
+            Files.createFile(jarFile);
+        }
 
         final Path sourcesFile = entry.asSources().buildArtifactPath(getRepositoryDirectory().get().getAsFile().toPath());
-        Files.deleteIfExists(sourcesFile);
-        Files.createFile(sourcesFile);
+        if (!Files.isRegularFile(sourcesFile)) {
+            Files.deleteIfExists(sourcesFile);
```
shartte: What is the point of deleteIfExists if it's not a file

shartte (changes requested):

common/src/main/java/net/neoforged/gradle/common/extensions/repository/IvyDummyRepositoryExtension.java
Matyrobbrt: It could be a directory.

shartte: Then deleteIfExists is only gonna do anything if it's empty :-P

PR #170, opened 2024-05-14
Create Conventions subsystem

marchermans:
### TLDR
This new subsystem (allthough only partially configurable via DSL), allows the management of NeoForge resources via conventions over configurations, runs and sourcesets.

The buildscript as such can be shorted to the following:
```groovy
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(21)
    }
}

repositories {
    mavenCentral()
}

dependencies {
    implementation 'net.neoforged:neoforge:+'
}
```
Obviously the version for neoforge can still be hand set.
This shorter variant will configure the same setup, including runs, configurations and all setup as usual. A reference to the neoforge implementation is still needed.

---

### Conventions
Within the gradle community there is a heavy emphasis on conventions, configurations are named in certain patterns, source sets trigger the creation of configurations, and main exists always, etc.

NeoGradle followed some of these, but not others. The goal of this PR is to follow these more, specifically when it comes to configurations, runs, and sourcesets.

This entire new subsystem is configurable like any other via the gradle properties, or the DSL.
> [!WARNING]
> Some properties of the new subsystem are only configurable via the gradle properties, as they are read as soon as the project is applied, others can be modified later, or when they are modified later, then the new value with only effect configurations, runs or sourcesets from the moment that it was set.

#### Configurations:
Per source set we now create the following configurations, using test as an example the name for the main sourceset can be found in brackets:
- testLocalRuntime (localRuntime) -> A configuration that exists so that you can add dependencies to the runtimeClasspath without them showing up externally in the runtimeElements of your sourceset or project.
- testLocalRunRuntime (localRunRuntime) -> A configuration that exists so that you can add dependencies to the runs runtime your sourceset is added to, as long as you have the relevant convention enabled (see below).

Per run we now also create the following configurations, using client as an example:
- clientRun -> A configuration to add dependencies to your runs runtime.

Globally we now also create 2 configurations:
- runs -> A configuration to add your dependency as a runtime element to any run.

#### Sourceset management:
We now add the main sourceset to any run in the project by default.
Additionally if you add a sourceset to the mod sources of a run, its localRunRuntime (so for example testLocalRunRuntime for test) configuration is also added to the runs runtime.

> [!NOTE]
> We currently do not offer a sourceset scoped mod configuration, but if the community requests one, we can trivially add it

#### Run management:
Automatic creation of runs based on available run types.
Most modders run with exactly one version of neoforge (either in one or more sourcesets), and they create all available runs; client, server, datagen, and gameTestServer. So this convention configures the runs accordingly, automatically creating a run based on the given run type found. If you run with multiple different NeoForge versions in different sourcesets, please dissable this convention via gradle.properties to prevent issues.
> [!NOTE]
> If you are effected by this, let us know so that we can gauge the community use and interest of the multi-neoforge version feature and possibly improve compatibilty.

#### IDE Integrations:
_This PR only offers conventions for the IDEA plugin:_
We now automatically detect if you are running NeoGradle in an IDEA instance with "Compile with IDEA" enabled. If this is the case different output paths are emitted.
We also now offer a direct convention property to set the output directory of the compiler. By default this is `out`. We do not yet automatically detect this from your configuration.

#### Properties
If you want to disable any of these functionalities, please see the new `readme.md` entries for them, or visit the NeoGradle documentation, once it is added there.

---

### Other changes:
- Due to modifications in the way the plugin initializes you can now configure the runs before the dependencies (usefull for if you want to use configuration based setups that add dependencies that way to the runs runtime or mods setup)
- NeoGradle will now always create a LocalRuntime configuration for all sourcesets, you do not need to create one yourself.


utils/src/main/java/net/neoforged/gradle/util/FileUtils.java
```diff
@@ -310,4 +310,14 @@ public static String postFixClassifier(final File file, final String classifier)
...
     }
+
+    public static boolean contains(final File file, final String shouldContain) {
+        try (final InputStream in = Files.newInputStream(file.toPath())) {
+            final byte[] bytes = new byte[in.available()];
+            in.read(bytes);
+            return new String(bytes).contains(shouldContain);
+        } catch (final IOException e) {
+            return false;
+        }
```
shartte: I'd go with Files.readAllBytes

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/runs/run/DependencyHandler.groovy
```diff
@@ -16,15 +13,30 @@ import org.gradle.api.tasks.Internal
...
+     * @return The dependency collector.
      */
+    @Internal
     DependencyCollector getRuntime();
+
+    /**
+     * Adds a dependency to the mods configuration.
+     *
+     * @return The dependency collector.
+     */
```
shartte: You should probably say what effect that has (i.e. placing them in the mods folder)

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/extensions/subsystems/conventions/Runs.groovy
```diff
@@ -0,0 +1,24 @@
...
+     * Note: this can not be configured via the buildscript and needs to be set in the gradle.properties file.
+     */
+    @DSLProperty
+    Property<Boolean> getIsEnabled()
+
+    /**
+     * Whether or not the default runs should be created.
+     */
+    @DSLProperty
+    Property<Boolean> getShouldDefaultRunsBeCreated()
```
shartte, line 23: How do these two differ from each other?

shartte (approved):

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/extensions/subsystems/conventions/Runs.groovy
marchermans: They don't for now, but each convention section has an enabled, right now there is obviously only one field here, and both need to be true, but that might not be so in the future.

PR #171, opened 2024-05-16
[Tech-Debt] Reduce complexity on Dependency Replacement and fix Bugs

marchermans:
### TLDR:
- Reduces complexity on Ivy Repository construction
- Reduces complexity on Dependency Replacement
- Reduced complexity on Runtime.
- Fixes several bugs related to how core components are registered
- IDEA Sync now happens immediately, eclipse sync will follow, once feedback is gathered
- Fixes to source based ATs and Parchment
- Reintroduce binary AT apply for Vanilla

---

### Ivy Repository refactor
The core change here is to remove all callbacks and create the core management files directly, additionally due to changes in the way dynamic dependencies work we do not need the `ng_dummy_ng` prefix anymore.

### Dynamic dependency replacement
The core change here, is to switch to a pure direct mode. Due to changes in the Ivy Repository we can now create all tasks and systems immediately on request and don't need to bother with afterEvaluate or even afterBake.

The introduction of a dedicated `ReplacementAware` interface allows for better notifications to flow through our system. Other subsystems can get notified of a replaced dependency happening, and what the replacement tasks are going to be.

### Reduced complexity on runtimes.
This is a first pass to reduce the complexity on runtimes, most of the "onBake" procedures are gone, and what is left can likely be further condensed and moved in the future.

### IDEA Sync
Using a new convention we can now run the IDEA sync immediately, no more post sync tasks, which should make switching between NG7 and NG6/FG6 a lot easier goting forward.

### Fixes:
- The raw jar for Userdev now is constructed properly again after it broke due to source ATs.
- Vanilla can now do ATs again, due to the reintroduction of binary AT application
- Registration order has been relaxed


common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -156,6 +147,24 @@ public void apply(Project project) {
...
+            @Override
+            public void idea(Project project, IdeaModel idea, ProjectSettings ideaExtension) {
+                setParallel(project);
+            }
+
+            private void setParallel(Project project) {
+                project.getGradle().getStartParameter().setParallelProjectExecutionEnabled(true);
+            }
+        });
+
```
shartte: The comment does not tell us why you are enforcing it.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -248,6 +258,46 @@ private void configureIDEAIDEConventions(Project project, IDE ideConventions) {
...
+                        } else {
+                            taskPath = String.format("%s:%s", projectPath, ideImportTaskName);
+                        }
+                    }
+
+                    taskPaths.add(taskPath);
+
+                    taskRequests.add(new DefaultTaskExecutionRequest(taskPaths));
+                    startParameter.setTaskRequests(taskRequests);
+                });
```
shartte, line 281: Are you certain this works?

shartte: I should clarify: you are modifying the start request, but repeated syncs may be running via the tooling API in the same daemon. Are you sure that it even uses distinct start requests in that scenario?

marchermans: Yes, it always has to, the tooling just starts a new gradle run. It does nothing special as far as I can see. Just wraps the execution of gradle and allows for data-model extraction.

marchermans: And yes this works, I tested it even on a multi-loader projects and this is the exact codepath Loom also uses to achieve the same thing.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
marchermans: Actually I need to rip that
Was testing start parameter management with this. It has no effect.

marchermans: Done

PR #173, opened 2024-05-16
[Tech-Debt] Switch to using full named configurations.

### LINKED ISSUE #136: Run Setup task issues when using different sourceset
Sorry, I couldn't come up with a short better title.

Say you have your mod source in the `main` SourceSet, but you want to run something else in another SourceSet,
in this case you can create a run configuration and make it's `xxxImplementation` configuration extend `implementation`

See below for an extremely minimalistic example (It crashes after start because there are no compiled classes, but that's not the point)

```
plugins {
    id 'java-library'
    id 'net.neoforged.gradle.userdev' version '7.0.97'
}

java.toolchain.languageVersion = JavaLanguageVersion.of(17)

sourceSets {

    modRun {
        java.setSrcDirs([])
        resources.setSrcDirs([])
    }

}

runs {
    client {
        modSource sourceSets.modRun
    }
}

configurations {
    modRunImplementation.extendsFrom implementation
}

dependencies {
    implementation "net.neoforged:neoforge:${neo_version}"
}
```

This will launch (and then crash as mentioned above) without issues.
However, if we move the `configurations` block below the dependencies block it will not launch .
Not because the dependencies aren't there, but because the following tasks aren't executed before run (The last one is the main issue)
`neoFormJoined_xxx_DownloadAssets` `neoFormJoined_xxx_ExtractNatives` `writeMinecraftClasspathClient`

From a quick look it looks like this is because neoforge detectes when `neoforge` is added as a dep and then does fancy stuff around it. This is of course not triggered if a configuration want to inherit another one after this happened.

Side point:
Currently you need to use 
```
configurations {
    modRunImplementation.extendsFrom implementation
}
```
to make `modRun` work as a valid launch target, but theoretically you'd want the following to be enough:
```
configurations {
    modRunRuntimeClassPath.extendsFrom runtimeClasspath
}
```

### LINKED ISSUE #144: Add handling for pom dependencies
When adding a dependency on something like <https://mvnrepository.com/artifact/org.graalvm.polyglot/python/23.1.2>, Gradle will resolve this to the `pom` and pull in it's dependencies transitively.

Resolving the dependency directly will result in the `.pom` file, `UnionFileSystem` expects entries to be a zip/jar/directory, adding the pom resulting in `ProviderNotFoundException`s in `UnionFileSystem#openFileSystem` at runtime.

[Originally from Discord](https://discord.com/channels/313125603924639766/801175194298744902/1229727626869084241)

### LINKED ISSUE #165: `Cloud not clean up target directory` typo'ed error message
Tiny typo I saw that was a bit funny to see.
image

Come from this line:
https://github.com/neoforged/NeoGradle/blob/2d1add35e5883b7b5694106c5573dee04bdf08ba/utils/src/main/java/net/neoforged/gradle/util/CopyingFileTreeVisitor.java#L56

Cloud be cleaned up as a small side thing in a future NeoGradle update.

### LINKED ISSUE #172: Mixin extra is missing on Neogradle 7.0.128
I cloned MDK(https://github.com/neoforged/MDK/tree/dad00053f00af616e8e26171383427f52988fd1f) and it now doesn't have mixin extra.
If I add this line to ExampleMod.java, it fails to compile:
```java
import com.llamalad7.mixinextras.sugar.Local;
```

Neogradle 7.0.127 doesn't have this issue

### LINKED ISSUE #174: NeoGradles recompile process does not create a valid Jar
We do not have the manifest as first or second entry

marchermans:
### TLDR:
- I was stupid.
- Well Gradle is stupid.

---

I will await the first bug report where people fuck with these configs but so be it.

Closes #172
Closes #136 
Closes #144 
Closes #174
Closes #165


userdev/src/functionalTest/groovy/net/neoforged/gradle/userdev/FunctionalTests.groovy
```diff
@@ -155,6 +193,74 @@ class FunctionalTests extends BuilderBasedTestSpecification {
...
+        given:
+        def project = create("userdev_supports_loading_from_buildcache", {
+            it.build("""
+            java {
+                toolchain {
+                    languageVersion = JavaLanguageVersion.of(21)
+                }
+            }
+            
+            //Note: This cannot be static so that fg.repository can be properly accessed
```
pupnewfster: This comment you copy pasted is no longer valid (and I forgot to remove it when switching from FG to NG). Also I think instead of this helper method as you only have one repo, you probably could get away with:
```groovy
repositories {
    exclusiveContent {
        forRepository {
            maven { url 'https://maven.tterrag.com/' }
        }
        filter { includeGroup('team.chisel.ctm') }
    }
}
```

PR #180, opened 2024-05-18
[Fix] Update JST and add back support for MultiSourceSet setups

marchermans:
- Ensure that multiple sourcesets in the same project are supported
- Recompile the game with the required java version, regardless of what the user configures!
- SATs now default to the user selected runtime, specs don't have a version json so we can not pull it from there.
- Add the ability to disable the process fork for the recompiler.

### DIFF
--- a/dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/Constants.groovy
+++ b/dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/Constants.groovy
@@ -27,7 +27,7 @@ class Constants {
     public static final String DEFAULT_PARCHMENT_GROUP = "org.parchmentmc.data"
     public static final String DEFAULT_PARCHMENT_ARTIFACT_PREFIX = "parchment-"
     public static final String DEFAULT_PARCHMENT_MAVEN_URL = "https://maven.parchmentmc.org/"
-    public static final String JST_TOOL_ARTIFACT = "net.neoforged.jst:jst-cli-bundle:1.0.35"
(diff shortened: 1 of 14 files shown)


dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/Constants.groovy
```diff
@@ -27,7 +27,7 @@ class Constants {
     public static final String DEFAULT_PARCHMENT_GROUP = "org.parchmentmc.data"
     public static final String DEFAULT_PARCHMENT_ARTIFACT_PREFIX = "parchment-"
     public static final String DEFAULT_PARCHMENT_MAVEN_URL = "https://maven.parchmentmc.org/"
-    public static final String JST_TOOL_ARTIFACT = "net.neoforged.jst:jst-cli-bundle:1.0.35"
+    public static final String JST_TOOL_ARTIFACT = "net.neoforged.jst:jst-cli-bundle:1.0.36"
```
Matyrobbrt, line 30: I would make this globally configurable instead of just in parchment.

marchermans: What do you mean?

Matyrobbrt: The Parchment subsystem can configure the JST artifact (and therefore version) used to apply parchment. The JST version used by ATs should also be configurable (mostly for testing).

PR #181, opened 2024-05-18
[Feature]: Introduce java version management

marchermans:
When creating a runtime, NG7 will automatically set the required java version for your game version.

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
@@ -24,6 +24,6 @@ public class JavaVersionManager {
         }
 
         this.javaVersion = javaVersion;
-        project.getExtensions().getByType(JavaPluginExtension.class).getToolchain().getLanguageVersion().set(JavaLanguageVersion.of(javaVersion));
+        project.getExtensions().getByType(JavaPluginExtension.class).getToolchain().getLanguageVersion().convention(JavaLanguageVersion.of(javaVersion));
     }
 }
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
@@ -1,0 +1,29 @@
+package net.neoforged.gradle.common.extensions;
+
+import org.gradle.api.Project;
+import org.gradle.api.plugins.JavaPluginExtension;
+import org.gradle.jvm.toolchain.JavaLanguageVersion;
+import org.gradle.jvm.toolchain.internal.JavaToolchain;
+
+import javax.inject.Inject;
+
+public class JavaVersionManager {
+
+    private final Project project;
+    private int javaVersion = -1;
+
+    @Inject
+    public JavaVersionManager(Project project) {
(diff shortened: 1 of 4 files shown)


common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
```diff
@@ -0,0 +1,29 @@
+package net.neoforged.gradle.common.extensions;
```
Matyrobbrt, line 1: Just.. set the version? What's the point of this extension?

Matyrobbrt (changes requested):

common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
```diff
@@ -0,0 +1,29 @@
...
+    private final Project project;
+    private int javaVersion = -1;
+
+    @Inject
+    public JavaVersionManager(Project project) {
+        this.project = project;
+    }
+
+    public void setJavaVersion(int javaVersion, String context) {
+        if (this.javaVersion > javaVersion) {
```
Matyrobbrt, line 21: The logic needs to be the other way around. In the case where there's 2 neo deps in the same project, the project must compile with the lowest Java version so that both sourcesets will work. Compiling with the highest will cause the one needing the lower to crash at runtime, and even if target compatibility were set, one can easily run into the issue of using newer Java methods.

Matyrobbrt (changes requested):

common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
```diff
@@ -0,0 +1,29 @@
...
+            project.getLogger().warn("Can not reconfigure java version from {} to {} for {}", this.javaVersion, javaVersion, context);
+            return;
+        }
+
+        this.javaVersion = javaVersion;
+        project.getExtensions().getByType(JavaPluginExtension.class).getToolchain().getLanguageVersion().set(JavaLanguageVersion.of(javaVersion));
```
lukebemish: Gotta agree with maty here, this seems rather pointless...
...but if you are going to use this, you should at least make this `convention` not `set` so that user-added java versions override this one. This shouldn't override what users set, it should just provide something so that they don't _have_ to set it.

common/src/main/java/net/neoforged/gradle/common/extensions/JavaVersionManager.java
lukebemish: The issue is that the toolchain is used for running too, and for running you _definitely_ need the newest. And NG doesn't currently have an easy way to use a separate javalauncher for a given run configuration it seems? Maybe the real solution would be to add that -- then the runs with a given MC version could use the one for that MC version and the root toolchain would use the oldest.

marchermans: Just to clarfy, not just running, also decompile and recompile need the newest versions, so this is the correct default behaviour.

marchermans:
Closing PR as this is a chicken and egg situation:
Need to resolve a configuration to download tools and the version json data.
This causes the java version to become final.

PR #192, opened 2024-05-29
[Feature]: Add support for DevLogin

### LINKED ISSUE #100: [Feature Request] Add DevLogin support to client run configs
[DevLogin](https://github.com/covers1624/DevLogin#devlogin) allows users to authenticate their clients using Mojangs services at startup.

Adding DevLogin to any NeoForge mod involves the [following steps currently](https://github.com/covers1624/DevLogin/wiki/NeoGradle).

While this is a simple setup, I think it would make sense for NeoGradle, and potentially other loaders too, to include DevLogin as a feature for their client run configs.
Of course this would have to be a opt-in feature, having it enabled by default wouldn't make much sense IMO.

I am thinking of adding a super simple boolean flag (`authed true`) to client configs, which when enabled
- sets the main class to `net.covers1624.devlogin.DevLogin`
- sets the program arg `--launch_target` to `cpw.mods.bootstraplauncher.BootstrapLauncher`
- and tells DevLogin to use the `default` auth profile via `--launch_profile default`
	There should also be an optional property users can specify to set the profile that DevLogin will use. This will allow users to authenticate their clients with multiple accounts.
	Maybe something like `authProfile <profile>` (if set also sets `authed true`)

marchermans:
### TLDR: 
- This adds supports for DevLogin on client runs

---

### DevLogin Support:
We now support the use of DevLogin by @covers1624.
You can use the following DSL to enable it:
```groovy
runs {
    someRun {
        devLogin {
           enabled true
        }
    }
}
```
By default the tool is disabled on all client runs.

> [!WARNING]
> If you enable the dev login tool for a none client run, you will get an error message.

### Better error handling:
There is now better error handling if you do not configure mod sources in a run.

### Closes:
Closes: #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,3 +359,57 @@ The following properties can be used to configure the JST tool:
 ```properties
 neogradle.subsystems.tools.jst=<artifact coordinate for jst cli tool>
 ```
+### DevLogin
+This tool is used by the runs subsystem to enable logged in plays on all client runs.
+The following properties can be used to configure the DevLogin tool:
+```properties
+neogradle.subsystems.tools.devLogin=<artifact coordinate for devlogin cli tool>
+```
+
+## DevLogin
+The DevLogin tool is a tool that allows you to log in to a Minecraft account without having to use the Minecraft launcher, during development.
+This tool is used by the runs subsystem to enable logged in plays on all client runs.
+The tool can be configured using the following properties:
+```properties
+neogradle.subsystems.devLogin.enabled=<true/false>
+```
+By default, the subsystem is enabled, and it will ask you to log in when you run a client run.
+If you want to disable this, you can set the property to false, and then you will not be asked to log in, but use a random none signed in dev account.
+
+Additionally, you can configure whether to add the relevant repositories to your buildscript:
+```properties
+neogradle.subsystems.devLogin.addRepositories=<true/false>
+``` 
+By default, the repositories are added, but if you want to disable this, you can set the property to false.
+By doing so you are responsible for adding a maven repository that provides the dev login CLI artifact.
+
+## Centralized Cache
+NeoGradle has a centralized cache that can be used to store the decompiled Minecraft sources, the recompiled Minecraft sources, and other task outputs of complex tasks.
+The cache is enabled by default, and can be disabled by setting the following property in your gradle.properties:
+```properties
+net.neoforged.gradle.caching.enabled=false
+```
+
+You can clean the artifacts that are stored in the cache by running the following command:
+```shell
+./gradlew cleanCache
+```
+
+This command is also automatically ran, when you run the clean task.
+The command will check if the stored artifact count is higher than the configured threshold, and if so remove the oldest artifacts until the count is below the threshold.
+The count is configured by the following property in your gradle.properties:
+```properties
+net.neoforged.gradle.caching.maxCacheSize=<number>
+```
+
+### Debugging
+There are two properties you can tweak to get more information about the cache:
+```properties
+net.neoforged.gradle.caching.logCacheHits=<true/false>
+```
+and
+```properties
+net.neoforged.gradle.caching.debug=<true/false>
+```
+The first property will log when a cache hit occurs, and the second property will log more information about the cache in general, including how hashes are calculated.
+If you are experiencing issues with the cache, you can enable these properties to get more information about what is happening.
--- a/README.md
+++ b/README.md
@@ -392,7 +392,7 @@ runs {
     }
 }
 ```
-This will disable the dev login tool for this run. By default, the dev login tool is enabled for all client runs.
+This will disable the dev login tool for this run. By default, the dev login tool is disable, and can only be enabled for client runs.
 
 > [!WARNING]
 > If you enable the dev login tool for a none client run, you will get an error message.
--- a/README.md
+++ b/README.md
@@ -383,6 +383,20 @@ neogradle.subsystems.devLogin.addRepositories=<true/false>
 By default, the repositories are added, but if you want to disable this, you can set the property to false.
 By doing so you are responsible for adding a maven repository that provides the dev login CLI artifact.
 
+### Per run configuration
+If you want to configure the dev login tool per run, you can do so by setting the following properties in your run configuration:
+```groovy
+runs {
+    someRun {
+        shouldUseDevLogin false
+    }
+}
+```
+This will disable the dev login tool for this run. By default, the dev login tool is enabled for all client runs.
+
+> [!WARNING]
+> If you enable the dev login tool for a none client run, you will get an error message.
+
 ## Centralized Cache
 NeoGradle has a centralized cache that can be used to store the decompiled Minecraft sources, the recompiled Minecraft sources, and other task outputs of complex tasks.
 The cache is enabled by default, and can be disabled by setting the following property in your gradle.properties:
--- a/README.md
+++ b/README.md
@@ -335,7 +335,6 @@ By default, the import in IDEA is run during the sync task.
 If you want to disable this, and use a post sync task, you can set the following property in your gradle.properties:
 ```properties
 neogradle.subsystems.conventions.ide.idea.use-post-sync-task=true
-
 ```
 
 ### Runs
@@ -365,6 +364,7 @@ The following properties can be used to configure the DevLogin tool:
 ```properties
 neogradle.subsystems.tools.devLogin=<artifact coordinate for devlogin cli tool>
 ```
+More information on the relevant tool, its released version and documentation can be found here: [DevLogin by Covers1624](https://github.com/covers1624/DevLogin)
 
 ## DevLogin
 The DevLogin tool is a tool that allows you to log in to a Minecraft account without having to use the Minecraft launcher, during development.
--- a/README.md
+++ b/README.md
@@ -383,7 +383,9 @@ If you want to configure the dev login tool per run, you can do so by setting th
 ```groovy
 runs {
     someRun {
-        useDevLogin true
+        devLogin {
+            enabled true
+        }
     }
 }
 ```
@@ -398,6 +400,18 @@ If you want to enable the dev login tool for all client runs, you can set the fo
 neogradle.subsystems.devLogin.conventionForRun=true
 ```
 This will enable the dev login tool for all client runs, unless explicitly disabled.
+
+Additionally, it is possible to use a different user profile for the dev login tool, by setting the following property in your run configuration:
+```groovy
+runs {
+    someRun {
+        devLogin {
+            profile '<profile>'
+        }
+    }
+}
+```
+If it is not set then the default profile will be used. See the DevLogin documentation for more information on profiles: [DevLogin by Covers1624](https://github.com/covers1624/DevLogin/blob/main/README.md#multiple-accounts)
 
 ### Configurations
 To add the dev login tool to your run we create a custom configuration to which we add the dev login tool.
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ The following properties can be used to configure the JST tool:
 neogradle.subsystems.tools.jst=<artifact coordinate for jst cli tool>
 ```
 ### DevLogin
-This tool is used by the runs subsystem to enable logged in plays on all client runs.
+This tool is used by the runs subsystem to enable Minecraft authentication in client runs.
 The following properties can be used to configure the DevLogin tool:
 ```properties
 neogradle.subsystems.tools.devLogin=<artifact coordinate for devlogin cli tool>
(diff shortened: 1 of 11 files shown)


Technici4n (dismissed):
Please make it disabled by default. At least in neodev.

marchermans:
> Please make it disabled by default. At least in neodev.

Is there a reason why?
Not against this, just want to know why, so I can leave a comment explaining the need to disable it in dynprojects.

marchermans:
> > Please make it disabled by default. At least in neodev.
> 
> Is there a reason why? Not against this, just want to know why, so I can leave a comment explaining the need to disable it in dynprojects.

Clarified via discord. Was personal preference of @Technici4n.

We decided to disable the flag on the run, this leaves the subsystem enabled, while having no effect for the user.

README.md
```diff
@@ -359,3 +359,71 @@ The following properties can be used to configure the JST tool:
 ```properties
 neogradle.subsystems.tools.jst=<artifact coordinate for jst cli tool>
 ```
+### DevLogin
+This tool is used by the runs subsystem to enable logged in plays on all client runs.
+The following properties can be used to configure the DevLogin tool:
+```properties
+neogradle.subsystems.tools.devLogin=<artifact coordinate for devlogin cli tool>
```
sciwhiz12, line 365: Perhaps this should document the GitHub repository of the DevLogin tool.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -374,6 +378,30 @@ private void applyAfterEvaluate(final Project project) {
...
+                            defaultRuntimeOnlyConfiguration.getDependencies().add(project.getDependencies().create(tools.getDevLogin().get()));
+
+                            //Update the program arguments to properly launch the dev login tool
+                            run.getProgramArguments().add("--launch_target");
+                            run.getProgramArguments().add(mainClass);
+
+                            //Set the main class to the dev login tool
+                            run.getMainClass().set("net.covers1624.devlogin.DevLogin");
+                        } else if (!runImpl.getIsClient().get() && runImpl.getShouldUseDevLogin().get()) {
+                            throw new GradleException("Dev login is only supported on runs which are marked as clients! The run: " + runImpl.getName() + " is not a client run.");
```
sciwhiz12: Perhaps [`InvalidUserDataException`](https://docs.gradle.org/8.7/javadoc/org/gradle/api/InvalidUserDataException.html) would be a more fitting exception here, than the generic `GradleException`.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -374,6 +378,30 @@ private void applyAfterEvaluate(final Project project) {
...
+                            final SourceSet defaultSourceSet = runImpl.getModSources().get().get(0);
+                            final Configuration defaultRuntimeOnlyConfiguration = project.getConfigurations().maybeCreate(defaultSourceSet.getRuntimeOnlyConfigurationName());
+                            defaultRuntimeOnlyConfiguration.getDependencies().add(project.getDependencies().create(tools.getDevLogin().get()));
+
+                            //Update the program arguments to properly launch the dev login tool
+                            run.getProgramArguments().add("--launch_target");
+                            run.getProgramArguments().add(mainClass);
+
+                            //Set the main class to the dev login tool
+                            run.getMainClass().set("net.covers1624.devlogin.DevLogin");
```
sciwhiz12: Is this configurable by the end-user? As in, if they switch out the DevLogin artifact in the future, can they change the main class of the run manually, or is this fixed in NeoGradle now?

README.md
```diff
@@ -359,3 +359,72 @@ The following properties can be used to configure the JST tool:
...
+``` 
+By default, the repositories are added, but if you want to disable this, you can set the property to false.
+By doing so you are responsible for adding a maven repository that provides the dev login CLI artifact.
+
+### Per run configuration
+If you want to configure the dev login tool per run, you can do so by setting the following properties in your run configuration:
+```groovy
+runs {
+    someRun {
+        shouldUseDevLogin true
```
sciwhiz12: Why is it called `shouldUseDevLogin`? Why not `useDevLogin`?

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/Constants.groovy
```diff
@@ -28,6 +28,11 @@ class Constants {
     public static final String DEFAULT_PARCHMENT_ARTIFACT_PREFIX = "parchment-"
     public static final String DEFAULT_PARCHMENT_MAVEN_URL = "https://maven.parchmentmc.org/"
     public static final String JST_TOOL_ARTIFACT = "net.neoforged.jst:jst-cli-bundle:1.0.36"
+    public static final String DEFAULT_DEVLOGIN_GROUP = "net.covers1624"
+    public static final String DEFAULT_DEVLOGIN_ARTIFACT = "DevLogin"
+    public static final String DEFAULT_DEVLOGIN_MAVEN_URL = "https://maven.covers1624.net/"
```
Technici4n: Isn't it already available via our maven?

sciwhiz12 (changes requested):

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
marchermans: I had it hardcoded, but your comment made me add a subsystem property to configure it,

README.md
```diff
@@ -359,3 +359,72 @@ The following properties can be used to configure the JST tool:
...
+    }
+}
+```
+This will enable the dev login tool for this run. 
+By default, the dev login tool is disabled, and can only be enabled for client runs.
+
+> [!WARNING]
+> If you enable the dev login tool for a none client run, you will get an error message.
```
Matyrobbrt: You need the `non` prefix, not the `none` pronoun.

marchermans:
Addressed all comments.

README.md
```diff
@@ -359,3 +358,73 @@ The following properties can be used to configure the JST tool:
...
+``` 
+By default, the repositories are added, but if you want to disable this, you can set the property to false.
+By doing so you are responsible for adding a maven repository that provides the dev login CLI artifact.
+
+### Per run configuration
+If you want to configure the dev login tool per run, you can do so by setting the following properties in your run configuration:
+```groovy
+runs {
+    someRun {
+        shouldUseDevLogin true
```
sciwhiz12: You forgot to update this line to match the new property `useDevLogin`.

sciwhiz12 (changes requested):

README.md
```diff
@@ -359,3 +358,73 @@ The following properties can be used to configure the JST tool:
...
+```properties
+net.neoforged.gradle.caching.enabled=false
+```
+
+You can clean the artifacts that are stored in the cache by running the following command:
+```shell
+./gradlew cleanCache
+```
+
+This command is also automatically ran, when you run the clean task.
```
Matyrobbrt: This sentence is in the passive voice, which requires a past participle verb (`run` being the past participle of `run`)

README.md
```diff
@@ -359,3 +358,73 @@ The following properties can be used to configure the JST tool:
...
+The command will check if the stored artifact count is higher than the configured threshold, and if so remove the oldest artifacts until the count is below the threshold.
```
Matyrobbrt:
```suggestion
The command will check if the stored artifact count is higher than the configured threshold, and if so, remove the oldest artifacts until the count is below the threshold.
```

README.md
```diff
@@ -359,3 +358,73 @@ The following properties can be used to configure the JST tool:
...
+More information on the relevant tool, its released version and documentation can be found here: [DevLogin by Covers1624](https://github.com/covers1624/DevLogin)
+
+## DevLogin
+The DevLogin tool is a tool that allows you to log in to a Minecraft account without having to use the Minecraft launcher, during development.
+This tool is used by the runs subsystem to enable logged in plays on all client runs.
+The tool can be configured using the following properties:
+```properties
+neogradle.subsystems.devLogin.enabled=<true/false>
+```
+By default, the subsystem is enabled, and it will ask you to log in when you run a client run.
```
Matyrobbrt: This is contradicting with the lines below. Is it, or is it not enabled?

README.md
```diff
@@ -359,3 +358,73 @@ The following properties can be used to configure the JST tool:
 ```properties
 neogradle.subsystems.tools.jst=<artifact coordinate for jst cli tool>
 ```
+### DevLogin
+This tool is used by the runs subsystem to enable logged in plays on all client runs.
```
Matyrobbrt:
```suggestion
This tool is used by the runs subsystem to enable Minecraft authentification in client runs.
```

README.md
```diff
@@ -359,3 +358,73 @@ The following properties can be used to configure the JST tool:
...
+
+## DevLogin
+The DevLogin tool is a tool that allows you to log in to a Minecraft account without having to use the Minecraft launcher, during development.
+This tool is used by the runs subsystem to enable logged in plays on all client runs.
+The tool can be configured using the following properties:
+```properties
+neogradle.subsystems.devLogin.enabled=<true/false>
+```
+By default, the subsystem is enabled, and it will ask you to log in when you run a client run.
+If you want to disable this, you can set the property to false, and then you will not be asked to log in, but use a random none signed in dev account.
```
Matyrobbrt: No comma after if clauses. Also, `non`, not `none`.

common/src/main/java/net/neoforged/gradle/common/extensions/subsystems/SubsystemsExtension.java
```diff
@@ -33,13 +29,27 @@ public SubsystemsExtension(Project project) {
         configureRecompilerDefaults();
         configureParchmentDefaults();
         configureToolsDefaults();
+        configureDevLoginDefaults();
+    }
+
+    private void configureDevLoginDefaults() {
+        DevLogin devLogin = getDevLogin();
+        devLogin.getEnabled().convention(
+                getBooleanProperty("devLogin.enabled").orElse(true)
```
Matyrobbrt, line 38: Wasn't it meant to be false by default?

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -374,6 +378,30 @@ private void applyAfterEvaluate(final Project project) {
...
+                    final DevLogin devLogin = project.getExtensions().getByType(Subsystems.class).getDevLogin();
+                    final Tools tools = project.getExtensions().getByType(Subsystems.class).getTools();
+                    if (devLogin.getEnabled().get()) {
+                        //Dev login is only supported on the client side
+                        if (runImpl.getIsClient().get() && runImpl.getUseDevLogin().get()) {
+                            final String mainClass = runImpl.getMainClass().get();
+
+                            //We add the dev login tool to the runtime only configuration, of the first source set, this should suffice
+                            final SourceSet defaultSourceSet = runImpl.getModSources().get().get(0);
+                            final Configuration defaultRuntimeOnlyConfiguration = project.getConfigurations().maybeCreate(defaultSourceSet.getRuntimeOnlyConfigurationName());
```
Matyrobbrt: `runtimeOnly` is a leaking configuration, please user another one that inherits from `runtimeClasspath`.

Matyrobbrt (dismissed):

common/src/main/java/net/neoforged/gradle/common/extensions/subsystems/SubsystemsExtension.java
marchermans: By lukes suggestion, and the one I agreed with, the subsystem it self is enabled, yet the default on each run is `false` so the user only needs to enable the relevant runs he or she wants.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
marchermans: Addressed, made a custom runtime only configuration and extended it in the same way that the localRuntime does when that is enabled.

README.md
```diff
@@ -359,3 +358,76 @@ The following properties can be used to configure the JST tool:
...
+### DevLogin
+This tool is used by the runs subsystem to enable Minecraft authentication in client runs.
+The following properties can be used to configure the DevLogin tool:
+```properties
+neogradle.subsystems.tools.devLogin=<artifact coordinate for devlogin cli tool>
+```
+More information on the relevant tool, its released version and documentation can be found here: [DevLogin by Covers1624](https://github.com/covers1624/DevLogin)
+
+## DevLogin
+The DevLogin tool is a tool that allows you to log in to a Minecraft account without having to use the Minecraft launcher, during development.
```
sciwhiz12, line 370: As I understand it, DevLogin works when the developer is logged into the vanilla Minecraft launcher, correct? That should be clarified here.

sciwhiz12 (changes requested):

README.md
```diff
@@ -359,3 +358,83 @@ The following properties can be used to configure the JST tool:
...
+By default, it will use the credentials stored by the vanilla launcher, however if you are not logged in there, it will offer you the ability to log in directly from the console.
```
covers1624: DevLogin does not have the ability to use the vanilla launcher credentials. It purely implements the device auth flow. If someone can figure out how they store the creds, we can absolutely add support for it.

covers1624:
Glad to see a tool I wrote to give me my skin in-dev put to real use.

It may be worth exposing DevLogin's profile system via the DSL `--launch_profile`, so modders can easily use their alts. Just a simple identifier string for the profile.

sciwhiz12 (approved):

PR #194, opened 2024-05-31
[Feature]: Improved testing

marchermans:
## Bugs fixed:
- We now properly resolve metadata of dependencies from mojangs maven.
- We now are able to lookup NeoForm mappings on windows 

## New Testing:
- This PR adds testing on MacOS, Windows and Linux separately
- This PR adds better handling for failure cases
- This PR adds fixes required for testing on different platforms

## New testing utils module:
- Contains groovy extensions used during testing

### DIFF
--- a/.github/gradle/gradle.gradle
+++ b/.github/gradle/gradle.gradle
@@ -83,18 +83,6 @@ class TestRun {
     String displayName
     String path
     String filter
-
-    String getDisplayName() {
-        return displayName
-    }
-
-    String getPath() {
-        return path
-    }
-
-    String getFilter() {
-        return filter
-    }
 }
 
 subprojects.forEach { Project subProject ->
--- a/.github/gradle/gradle.gradle
+++ b/.github/gradle/gradle.gradle
@@ -8,13 +8,12 @@ if (!project.hasProperty('githubCiTesting') && System.env['GITHUB_OUTPUT'] == nu
 
 static Map<String, String> removeCommonPrefix(List<String> list) {
     if (list.size() == 0) {
-        return new HashMap<String, String>()
+        return [:]
     }
 
     def result = new HashMap<String, String>()
 
-    if (list.size() == 1) {
-        System.out.println("Only one item in the list: ${list}")
+    if (list.unique(false).size() == 1) {
         def sections = list[0].split("\\.")
         result.put(list[0], sections.last())
         return result
@@ -23,21 +22,27 @@ static Map<String, String> removeCommonPrefix(List<String> list) {
     Collections.sort(list, { String a, String b -> a.length() - b.length() })
     def max = list[0].length()
     def prefix = ""
+
+    comparisonLoop:
     for(int index = 0; index < max; index++) {
         def currentChar = list[0][index]
         for (String item : list) {
             if (item[index] != currentChar) {
-                for (final def entry in list) {
-                    result.put(entry, entry.substring(prefix.length()))
-                }
-                return result
+                break comparisonLoop
             }
         }
 
         prefix += currentChar
     }
 
-    throw new RuntimeException("Could not remove common prefix: ${list}")
+    if (prefix.length() == max) {
+        throw new IllegalArgumentException("The common prefix is the entire string")
+    }
+
+    for (final def entry in list) {
+        result.put(entry, entry.substring(prefix.length()))
+    }
+    return result
 }
 
 static String createDisplayNameSuffix(String path, String filter) {
@@ -64,7 +69,7 @@ static String createDisplayNameSuffix(String path, String filter) {
 }
 
 static List<TestRun> createTestRuns(Task it, List<String> testClasses) {
-    def testRuns = new ArrayList<TestRun>()
+    def testRuns = []
     if (testClasses.size() == 0) {
         testRuns.add(new TestRun(displayName: "Test - ${createDisplayNameSuffix(it.path, null)}", path: it.path, filter: null))
         return testRuns
--- a/.github/gradle/gradle.gradle
+++ b/.github/gradle/gradle.gradle
@@ -1,18 +1,106 @@
+import groovy.json.JsonBuilder
+import groovy.transform.CompileStatic
+
 if (!project.hasProperty('githubCiTesting') && System.env['GITHUB_OUTPUT'] == null) {
     // If the property is not set and we are not running in the CI environment then we will do nothing
     return
 }
 
-static def outputUnitTest(Project project, Task test) {
-    project.getLogger().lifecycle("<< TEST >>${test.path}")
+static Map<String, String> removeCommonPrefix(List<String> list) {
+    if (list.size() == 0) {
+        return new HashMap<String, String>()
+    }
+
+    def result = new HashMap<String, String>()
+
+    if (list.size() == 1) {
+        System.out.println("Only one item in the list: ${list}")
+        def sections = list[0].split("\\.")
+        result.put(list[0], sections.last())
+        return result
+    }
+
+    Collections.sort(list, { String a, String b -> a.length() - b.length() })
+    def max = list[0].length()
+    def prefix = ""
+    for(int index = 0; index < max; index++) {
+        def currentChar = list[0][index]
+        for (String item : list) {
+            if (item[index] != currentChar) {
+                for (final def entry in list) {
+                    result.put(entry, entry.substring(prefix.length()))
+                }
+                return result
+            }
+        }
+
+        prefix += currentChar
+    }
+
+    throw new RuntimeException("Could not remove common prefix: ${list}")
 }
 
-static def outputClassTest(Project project, Task test, String className) {
-    project.getLogger().lifecycle("<< TEST >>${test.path} --tests \"${className}\"")
+static String createDisplayNameSuffix(String path, String filter) {
+    //The path starts with a : and then the project name:
+    path = path.substring(path.indexOf(":") + 1)
+
+    //Now split the path into parts
+    def parts = path.split(":")
+    def capitalizedParts = parts.collect { it.capitalize() }
+
+    //Combine the parts with ->
+    path = capitalizedParts.join(" -> ")
+
+    if (filter === null) {
+        return path
+    }
+
+    //Next the filter has its common prefix stripped, is however still in domain name form
+    def filterParts = filter.split("\\.")
+    def capitalizedFilterParts = filterParts.collect { it.capitalize() }
+    filter = capitalizedFilterParts.join("/")
+
+    return "${path} (${filter})"
+}
+
+static List<TestRun> createTestRuns(Task it, List<String> testClasses) {
+    def testRuns = new ArrayList<TestRun>()
+    if (testClasses.size() == 0) {
+        testRuns.add(new TestRun(displayName: "Test - ${createDisplayNameSuffix(it.path, null)}", path: it.path, filter: null))
+        return testRuns
+    }
+
+    def testClassesWithoutCommonPrefix = removeCommonPrefix(testClasses)
+    testClassesWithoutCommonPrefix.forEach { className, displayName ->
+        testRuns.add(new TestRun(displayName: "Test - ${createDisplayNameSuffix(it.path, displayName)}", path: it.path, filter: className))
+    }
+
+    return testRuns
+}
+
+@CompileStatic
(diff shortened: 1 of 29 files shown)


.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,131 @@
...
+        return displayName
+    }
+
+    String getPath() {
+        return path
+    }
+
+    String getFilter() {
+        return filter
+    }
```
Matyrobbrt: This is groovy, you already have getters generated, which are used by default.

test-utils/src/main/groovy/net/neoforged/gradle/utils/test/extensions/FileExtensions.groovy
```diff
@@ -0,0 +1,14 @@
+package net.neoforged.gradle.utils.test.extensions
+
+/**
+ * Extensions for the [File] class.
+ */
+class FileExtensions {
```
Matyrobbrt, line 8: Missing compilestatic

test-utils/src/main/groovy/net/neoforged/gradle/utils/test/extensions/RuntimeBuilderExtensions.groovy
```diff
@@ -0,0 +1,23 @@
+package net.neoforged.gradle.utils.test.extensions
+
+import net.neoforged.gradle.common.caching.CentralCacheService
+import net.neoforged.trainingwheels.gradle.functional.builder.Runtime
+
+/**
+ * Extensions for the [Runtime.Builder] class.
+ */
+class RuntimeBuilderExtensions {
```
Matyrobbrt, line 10: Missing compilestatic

test-utils/src/main/groovy/net/neoforged/gradle/utils/test/extensions/RuntimeBuilderExtensions.groovy
```diff
@@ -0,0 +1,23 @@
+package net.neoforged.gradle.utils.test.extensions
+
+import net.neoforged.gradle.common.caching.CentralCacheService
+import net.neoforged.trainingwheels.gradle.functional.builder.Runtime
+
+/**
+ * Extensions for the [Runtime.Builder] class.
```
Matyrobbrt: This isn't a valid link, and neither is the one in the javadoc below.

.github/workflows/test-prs.yml
```diff
@@ -83,45 +89,77 @@ jobs:
...
+          Write-Output "Determined name to be $NAME-${{ matrix.os }}"
+          if ([string]::IsNullOrEmpty($env:GITHUB_OUTPUT)) {
+            # We do not have github output, then use the set output command
+            Write-Output "::set-output name=artifact-name::$NAME-${{ matrix.os }}"
+            exit
+          }
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "artifact-name=$NAME-${{ matrix.os }}"
+
+      - name: Format test run name as artifact name
+        if: (success() || failure()) &&  runner.os != 'Windows'
```
Matyrobbrt, line 126: While Mac is a *nix system, if you're going to test on it explicitly (which will always take longer considering concurrent runners are limited to 5), it shouldn't be put in the same bucket as Linux.

But I'd prefer if we hadn't tested mac to be honest. It's less used than Windows and Linux (ci), and it's meant to act similarly to Linux for the most part.

Matyrobbrt (dismissed):

.github/gradle/gradle.gradle
marchermans: Sadly that does not actually suffice for the JsonBuilder included in groovy
I tested it without cause I thought it would do it, but then it does not include the properties in the json what so ever.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -116,7 +116,7 @@ public void apply(Project project) {
 
         project.getRepositories().maven(e -> {
             e.setUrl(UrlConstants.MOJANG_MAVEN);
-            e.metadataSources(MavenArtifactRepository.MetadataSources::artifact);
+            e.metadataSources(MavenArtifactRepository.MetadataSources::mavenPom);
```
sciwhiz12, line 119: This is the fix for the whole problem with `libraries.minecraft.net` containing artifacts that aren't present on Maven Central (for GAV coordinates that are also present on Central), correct?

This should be mentioned in the PR description at minimum, and possibly in the final squash commit's message.

.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,131 @@
...
 if (!project.hasProperty('githubCiTesting') && System.env['GITHUB_OUTPUT'] == null) {
     // If the property is not set and we are not running in the CI environment then we will do nothing
     return
 }
 
-static def outputUnitTest(Project project, Task test) {
-    project.getLogger().lifecycle("<< TEST >>${test.path}")
+static Map<String, String> removeCommonPrefix(List<String> list) {
+    if (list.size() == 0) {
+        return new HashMap<String, String>()
```
sciwhiz12: Maps in Groovy can be constructed with `[:]`, if I recall correctly.

.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,131 @@
...
+    }
+
+    def result = new HashMap<String, String>()
+
+    if (list.size() == 1) {
+        System.out.println("Only one item in the list: ${list}")
```
sciwhiz12: Leftover debugging statement?

.github/workflows/test-prs.yml
marchermans: This has nothing to do with mac vs linux. But with bash vs powershell......

And I think we should test on mac.
It is an OS a significant part of our modders use, and we do not actually use the runners for anything else either.

.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,131 @@
...
+    def prefix = ""
+    for(int index = 0; index < max; index++) {
+        def currentChar = list[0][index]
+        for (String item : list) {
+            if (item[index] != currentChar) {
+                for (final def entry in list) {
+                    result.put(entry, entry.substring(prefix.length()))
+                }
+                return result
+            }
```
sciwhiz12, line 34: For clarity, I think you could refactor this a bit by making this `if` condition cause a `break` on the outermost loop, and then moving the final iteration that does the prefix-stripping to after the loop.

.github/gradle/gradle.gradle
marchermans: Seems to work now...
Not sure what I fucked up when I thought I needed to add them

.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,131 @@
...
+        }
+
+        prefix += currentChar
+    }
+
+    throw new RuntimeException("Could not remove common prefix: ${list}")
```
sciwhiz12: if I understand this correctly, there are two cases where we might be unable to strip a common prefix:

1. All strings in the list had no common prefix, so `prefix` is empty; or
2. All strings in the list are the exact same, where `prefix` is non-empty.

In all other cases, there is a common prefix among unique strings.

If we follow the suggestion in my other review comment to move the `if` condition outside of the loop, we can gate this `RuntimeException` (which should really be an `IllegalArgumentException` I think) behind an `if (prefix.isEmpty() || list.unique(false).size() == 1)`

.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,119 @@
...
+    //Next the filter has its common prefix stripped, is however still in domain name form
+    def filterParts = filter.split("\\.")
+    def capitalizedFilterParts = filterParts.collect { it.capitalize() }
+    filter = capitalizedFilterParts.join("/")
+
+    return "${path} (${filter})"
+}
+
+static List<TestRun> createTestRuns(Task it, List<String> testClasses) {
+    def testRuns = new ArrayList<TestRun>()
```
sciwhiz12: Lists in Groovy can be constructed with `[]`, if I recall correctly.

.github/workflows/test-prs.yml
```diff
@@ -133,16 +171,29 @@ jobs:
...
           name: test-results
           path: junit.xml
+
+      - name: Failed build detection
+        uses: actions/github-script@v7
+        if: >-
+          ${{
+               contains(needs.*.result, 'failure')
+            || contains(needs.*.result, 'cancelled')
+            || contains(needs.*.result, 'skipped')
```
sciwhiz12, line 194: Do we think that skipped or cancelled tests are automatic test failures?

sciwhiz12 (changes requested):

.github/workflows/test-prs.yml
marchermans: Yes in this case yeah.
If not all tests complete successfully that means that we have an issue.

.github/gradle/gradle.gradle
```diff
@@ -23,21 +22,27 @@ static Map<String, String> removeCommonPrefix(List<String> list) {
...
+    comparisonLoop:
     for(int index = 0; index < max; index++) {
         def currentChar = list[0][index]
         for (String item : list) {
             if (item[index] != currentChar) {
-                for (final def entry in list) {
-                    result.put(entry, entry.substring(prefix.length()))
-                }
-                return result
+                break comparisonLoop
```
sciwhiz12:
```
Script '/home/runner/work/NeoGradle/NeoGradle/.github/gradle/gradle.gradle' line: 27

* What went wrong:
Could not compile script '/home/runner/work/NeoGradle/NeoGradle/.github/gradle/gradle.gradle'.
> startup failed:
  script '/home/runner/work/NeoGradle/NeoGradle/.github/gradle/gradle.gradle': 27: Statement labels may not be used in build scripts.
  In case you tried to configure a property named 'comparisonLoop', replace ':' with '=' or ' ', otherwise it will not have the desired effect.
   @ line 27, column 5.
         for(int index = 0; index < max; index++) {
         ^
  
  1 error
```
Quite disappointing. Oh well, revert my suggestions for this.

.github/gradle/gradle.gradle
```diff
@@ -8,13 +8,12 @@ if (!project.hasProperty('githubCiTesting') && System.env['GITHUB_OUTPUT'] == nu
 
 static Map<String, String> removeCommonPrefix(List<String> list) {
     if (list.size() == 0) {
-        return new HashMap<String, String>()
+        return [:]
     }
 
     def result = new HashMap<String, String>()
```
sciwhiz12, line 14: You should replace this with `[:]` as well.

.github/gradle/gradle.gradle
```diff
@@ -1,41 +1,124 @@
...
+@CompileStatic
+class TestRun {
+    String displayName
+    String path
+    String filter
 }
 
 subprojects.forEach { Project subProject ->
     subProject.tasks.register('determineTests') { Task it ->
+        def tests = new ArrayList<TestRun>()
```
sciwhiz12: Use `[]` here, as well as for `testClasses` below.

sciwhiz12 (changes requested):

.github/gradle/gradle.gradle
marchermans: Doing that with a def creates a weird map, that is not typed -.-
So I decided this is the better option.

test-utils/src/main/groovy/net/neoforged/gradle/utils/test/extensions/FileExtensions.groovy
marchermans: Can't be compilestatic, causes compiler errors in consumers.

test-utils/src/main/groovy/net/neoforged/gradle/utils/test/extensions/RuntimeBuilderExtensions.groovy
marchermans: Can't be compile static, causes compiler errors in consumers.

sciwhiz12 (dismissed):

sciwhiz12 (dismissed):

sciwhiz12 (approved):

PR #211, opened 2024-06-17
[Fix]: Fix ATs not applying.

### LINKED ISSUE #197: DevLogin conventionForRun applies to all Runs not just client Runs
NG Version: 7.0.139

When attempting to use: `neogradle.subsystems.devLogin.conventionForRun=true` it enables `devLogin` for all runs, not just client runs. And then it fails to import due to attempting to enable it for non client runs.

The [docs](https://github.com/neoforged/NeoGradle?tab=readme-ov-file#devlogin-1) state: "This will enable the dev login tool for all client runs, unless explicitly disabled."

But looking at [the code](https://github.com/neoforged/NeoGradle/blob/NG_7.0/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java#L156) if the flag is set, then it gets enabled for all runs, without filtering to only apply to client runs.

### LINKED ISSUE #201: Building with `org.gradle.parallel=true` fails due to resolving dependencies in other subprojects
Attempting to use recent versions of NeoGradle with parallel builds and multiple projects fails with the following error:

```
* What went wrong:
A problem occurred configuring root project 'mdk-example'.
> Could not determine the dependencies of null.
   > Current thread does not hold the state lock for project ':another'
```

This is a relatively recent regression — 7.0.133 works, while 7.0.137 fails — so I suspect this was introduced by #191.

## Steps to reproduce
 - Clone the MDK (tested at bfc7eb87a59b671a274dbb8d94aab28784beea8e).
 - Add a new subproject (does not need to have NeoGradle installed), and a dependency on that project. I've included the full patch below.
 - Run `./gradlew tasks --parallel` to configure the project.

<details><summary>Changes.patch</summary>

```diff
diff --git a/another/build.gradle b/another/build.gradle
new file mode 100644
index 0000000..8c234a6
--- /dev/null
+++ b/another/build.gradle
@@ -0,0 +1,7 @@
+plugins {
+  id("java-library")
+}
+
+dependencies {
+ implementation "com.google.guava:guava:31.1-jre"
+}
diff --git a/build.gradle b/build.gradle
index 700db60..4e48e4d 100644
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ plugins {
     id 'eclipse'
     id 'idea'
     id 'maven-publish'
-    id 'net.neoforged.gradle.userdev' version '7.0.133'
+    id 'net.neoforged.gradle.userdev' version '7.0.139'
 }
 
 version = mod_version
@@ -88,6 +88,7 @@ dependencies {
     // And its provides the option to then use net.minecraft as the group, and one of; client, server or joined as the module name, plus the game version as version.
     // For all intends and purposes: You can treat this dependency as if it is a normal library you would use.
     implementation "net.neoforged:neoforge:${neo_version}"
+    implementation project(":another")
 
     // Example optional mod dependency with JEI
     // The JEI API is declared for compile time use, while the full JEI artifact is used at runtime
diff --git a/gradle.properties b/gradle.properties
index a4a8575..16eacae 100644
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,6 +2,7 @@
 org.gradle.jvmargs=-Xmx1G
 org.gradle.daemon=false
 org.gradle.debug=false
+org.gradle.parallel=true
 
 #read more on this at https://github.com/neoforged/NeoGradle/blob/NG_7.0/README.md#apply-parchment-mappings
 # you can also find the latest versions at: https://parchmentmc.org/docs/getting-started
diff --git a/settings.gradle b/settings.gradle
index ada876e..4434e83 100644
--- a/settings.gradle
+++ b/settings.gradle
@@ -9,3 +9,5 @@ pluginManagement {
 plugins {
     id 'org.gradle.toolchains.foojay-resolver-convention' version '0.8.0'
 }
+
+include(":another")
```

</details>

[Full log output](https://gist.github.com/SquidDev/3174620d600c00fe3343768456be56f4)

### LINKED ISSUE #207: neoFormJoined[...]DownloadAssets task error in multiprojects
When I try `./gradlew build` on my project, I get this error:

`Reason: Task ':NeoForgeApi:neoFormJoined1.20.6-20240429.153634DownloadAssets' uses this output of task ':NeoForge:neoFormJoined1.20.6-20240429.153634DownloadAssets' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
`

full log: https://gist.github.com/mezz/b20871b078635a5b100ba49182a4dfd3
my current code is here: https://github.com/mezz/JustEnoughItems/pull/3529 

my understanding is that my `:NeoForgeApi` project is using output from the same-named task in `:NeoForge`, but it is not declared explicitly

this is pretty concerning, because `:NeoForgeApi` does not depend on the `:NeoForge` project at all
let me know if there's any additional info I can provide to help

### LINKED ISSUE #209: Missing specified Locales
One of the testers for Mekanism has a computer set to Turkish. They were having issues importing and building the project with the following error: https://gist.github.com/pupnewfster/a58642d1fa7fe26247e853be51c78869

Since then they have reinstalled windows setting the language to English and it works. I believe this is because there are various places in NeoGradle that use `toLowerCase` and `toUpperCase` without specifying a `Locale`. I would make a PR (and still can), but I am unsure whether I should be using `Locale.ROOT` or `Locale.ENGLISH` given the places in NeoGradle that do make use of specifying a `Locale` use a mixture of the two.

marchermans:
Fix ATs not applying.
Fix Locale handling in string processing in as many places as I reasonably can.
Fix Parallel running
Fix Parallel central cache locking
Fix recompile not being cached
Fix DownloadAssets causing issues in multi project runs
Fix DevLogin conventionForRun triggering issues on none Client runs by convention

Closes #209
Closes #207
Closes #201
Closes #197

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
+++ b/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
@@ -390,8 +390,6 @@ public class CommonProjectPlugin implements Plugin<Project> {
                         }
                     });
 
-
-
                     definitionSet.forEach((identifier, definition) -> {
                         definition.configureRun(runImpl);
                     });
--- a/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
+++ b/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
@@ -57,7 +57,6 @@ import org.gradle.plugins.ide.idea.IdeaPlugin;
 import org.jetbrains.gradle.ext.IdeaExtPlugin;
 
 import java.util.*;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 public class CommonProjectPlugin implements Plugin<Project> {
 
--- a/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
+++ b/common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
@@ -57,6 +57,7 @@ import org.gradle.plugins.ide.idea.IdeaPlugin;
 import org.jetbrains.gradle.ext.IdeaExtPlugin;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class CommonProjectPlugin implements Plugin<Project> {
 
@@ -159,7 +160,7 @@ public class CommonProjectPlugin implements Plugin<Project> {
         if (devLogin.getEnabled().get()) {
             runs.configureEach(run -> {
                 final RunDevLogin runsDevLogin = run.getExtensions().create("devLogin", RunDevLogin.class);
-                runsDevLogin.getIsEnabled().convention(devLogin.getConventionForRun());
+                runsDevLogin.getIsEnabled().convention(devLogin.getConventionForRun().zip(run.getIsClient(), (conventionForRun, isClient) -> conventionForRun && isClient));
             });
         }
 
@@ -388,6 +389,8 @@ public class CommonProjectPlugin implements Plugin<Project> {
                             throw new RuntimeException("Failed to configure run: " + run.getName() + " there are multiple runtime definitions found for the source set: " + sourceSet.getName(), e);
                         }
                     });
+
+
 
                     definitionSet.forEach((identifier, definition) -> {
                         definition.configureRun(runImpl);
--- a/common/src/main/java/net/neoforged/gradle/common/runtime/tasks/action/DownloadFileAction.java
+++ b/common/src/main/java/net/neoforged/gradle/common/runtime/tasks/action/DownloadFileAction.java
@@ -106,30 +106,25 @@ public abstract class DownloadFileAction implements WorkAction<DownloadFileActio
         }
     }
 
-    private static final class Monitor implements CopyProgressListener {
-        private final GradleInternalUtils.ProgressLoggerWrapper progress;
-
-        private Monitor(GradleInternalUtils.ProgressLoggerWrapper progress) {
-            this.progress = progress;
-        }
+    private record Monitor(GradleInternalUtils.ProgressLoggerWrapper progress) implements CopyProgressListener {
 
         @Override
-        public void start(CopyProgressEvent evt) {
-            progress.started();
-            progress.incrementDownloadProgress(evt.getReadBytes());
-        }
+            public void start(CopyProgressEvent evt) {
+                progress.started();
+                progress.incrementDownloadProgress(evt.getReadBytes());
+            }
 
-        @Override
-        public void progress(CopyProgressEvent evt) {
-            progress.incrementDownloadProgress(evt.getReadBytes());
-        }
+            @Override
+            public void progress(CopyProgressEvent evt) {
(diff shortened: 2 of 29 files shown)


common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -389,6 +390,8 @@ private void applyAfterEvaluate(final Project project) {
                         }
                     });
 
+
+
```
sciwhiz12: Blank lines?

common/src/main/java/net/neoforged/gradle/common/runtime/tasks/action/DownloadFileAction.java
```diff
@@ -106,30 +106,25 @@ public void execute() {
...
-        public void end(CopyProgressEvent evt) {
-            progress.incrementDownloadProgress(evt.getReadBytes());
-            progress.completed();
+            @Override
+            public void end(CopyProgressEvent evt) {
+                progress.incrementDownloadProgress(evt.getReadBytes());
+                progress.completed();
+            }
         }
-    }
```
sciwhiz12: It seems you over-indented this by one indent.

platform/src/main/java/net/neoforged/gradle/platform/extensions/DynamicProjectExtension.java
```diff
@@ -770,7 +770,7 @@ private void configureRunType(final Project project, final RunType runType, fina
 
         runType.getClasspath().from(runtimeClasspath);
 
-        Provider<String> assetsDir = runtimeDefinition.getAssets().flatMap(DownloadAssets::getAssetsDirectory).map(Directory::getAsFile).map(File::getAbsolutePath);
+        Provider<String> assetsDir = DownloadAssets.getAssetsDirectory(project, project.provider(runtimeDefinition::getVersionJson)).map(Directory::getAsFile).map(File::getAbsolutePath);  // runtimeDefinition.getAssets().flatMap().map(Directory::getAsFile).map(File::getAbsolutePath);
```
sciwhiz12: Leftover comment.

sciwhiz12 (changes requested):

common/src/main/java/net/neoforged/gradle/common/runtime/tasks/action/DownloadFileAction.java
marchermans: Reformatted. Should fix it. I changed some of this to a record.

common/src/main/java/net/neoforged/gradle/common/CommonProjectPlugin.java
```diff
@@ -57,6 +57,7 @@
 import org.jetbrains.gradle.ext.IdeaExtPlugin;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
```
sciwhiz12: I assume this is an unused import from earlier testing.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #213, opened 2024-06-20
[FIx]: Fix AT issues and mute execute output further

marchermans:
This fixes the issue where the JST AT applier would not get the right classpath to analyse the source with.
Now all dependencies of a given runtime are passed in into the JST engine, including the libraries list.

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/runtime/tasks/SourceAccessTransformer.java
+++ b/common/src/main/java/net/neoforged/gradle/common/runtime/tasks/SourceAccessTransformer.java
@@ -6,12 +6,7 @@ import net.neoforged.gradle.dsl.common.extensions.subsystems.Subsystems;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.plugins.JavaPluginExtension;
-import org.gradle.api.tasks.CacheableTask;
-import org.gradle.api.tasks.InputFile;
-import org.gradle.api.tasks.InputFiles;
-import org.gradle.api.tasks.PathSensitive;
-import org.gradle.api.tasks.PathSensitivity;
-import org.gradle.api.tasks.SkipWhenEmpty;
+import org.gradle.api.tasks.*;
 
 import java.io.File;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


marchermans:
Test are failing due to new FML format. Can be ignored.

common/src/main/java/net/neoforged/gradle/common/runtime/tasks/SourceAccessTransformer.java
```diff
@@ -38,6 +33,15 @@ public SourceAccessTransformer() {
...
                             args.add("--libraries-list=" + getLibraries().get().getAsFile().getAbsolutePath());
 
+                            final StringBuilder builder = new StringBuilder();
+                            getClasspath().forEach(f -> {
+                                if (!builder.isEmpty()) {
+                                    builder.append(File.pathSeparator);
+                                }
+                                builder.append(f.getAbsolutePath());
+                            });
+                            args.add("--classpath=" + builder.toString());
```
sciwhiz12, line 43: Do we think the classpath will ever become so large as to exceed OS limitations on command-line length (if any)?

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/Constants.groovy
```diff
@@ -27,7 +27,7 @@ class Constants {
     public static final String DEFAULT_PARCHMENT_GROUP = "org.parchmentmc.data"
     public static final String DEFAULT_PARCHMENT_ARTIFACT_PREFIX = "parchment-"
     public static final String DEFAULT_PARCHMENT_MAVEN_URL = "https://maven.parchmentmc.org/"
-    public static final String JST_TOOL_ARTIFACT = "net.neoforged.jst:jst-cli-bundle:1.0.38"
+    public static final String JST_TOOL_ARTIFACT = "net.neoforged.jst:jst-cli-bundle:1.0.39"
```
sciwhiz12, line 30: Latest JavaSourceTransformer version is `1.0.42`.

sciwhiz12 (commented):
I don't see any glaring errors in the logic (and I'm not able to thoroughly test). Just a handful of nitpicks.

Some of the tests should be fixed soon:tm:, particularly those tests which check for `SUCCESS` but see `UP_TO_DATE`. There's also some tests which should be looked at soon:tm: that involve the central cache service.

common/src/main/java/net/neoforged/gradle/common/runtime/tasks/SourceAccessTransformer.java
marchermans: Maybe, I have plans to fix this though :dagger: 
Just not in a PR that is a fix for an issue.

dsl/common/src/main/groovy/net/neoforged/gradle/dsl/common/util/Constants.groovy
marchermans: Does not matter for my case, will bump that in a later PR

PR #216, opened 2024-06-24
[Fix]: Feature/primary sourceset selector

### LINKED ISSUE #198: Configuration Cache Problems
NG Version: 7.0.139

It seems that when the configuration cache is enabled the following tasks break:
- `clean`: https://gist.github.com/pupnewfster/261ad95f5de5dec508431175a5e73c22
- `runClient`: https://gist.github.com/pupnewfster/16f37f8ca68ecef4e9a6208e8e218ff4

Buildscript: https://github.com/mekanism/Mekanism/blob/1.20.6/build.gradle

### LINKED ISSUE #214: Multiple copies of Minecraft on the runtime classpath (NG7+MDG issue)
Per `#modder-support-gradle`, "That looks like MDG is leaking its results into the runtime classpath"

```
Exception in thread "main" net.neoforged.fml.ModLoadingException: Loading errors encountered:
	- Found multiple copies of net/minecraft/client/Minecraft.class on the classpath: [F:\minecraft\compactmods\compactmachines\neoforge-1.20.x\neoforge-datagen\.gradle\repositories\ng_dummy_ng\net\neoforged\neoforge\21.0.14-beta\neoforge-21.0.14-beta.jar, F:\minecraft\compactmods\compactmachines\neoforge-1.20.x\neoforge-main\build\moddev\artifacts\neoforge-minecraft-joined-local-merged.jar, F:\minecraft\compactmods\compactmachines\neoforge-1.20.x\core-api\build\moddev\artifacts\neoforge-minecraft-joined-local-merged.jar]

	at MC-BOOTSTRAP/fml_loader@4.0.6/net.neoforged.fml.util.DevEnvUtils.findFileSystemRootOfFileOnClasspath(DevEnvUtils.java:78)
	at MC-BOOTSTRAP/fml_loader@4.0.6/net.neoforged.fml.loading.targets.CommonUserdevLaunchHandler.collectAdditionalModFileLocators(CommonUserdevLaunchHandler.java:24)
	at MC-BOOTSTRAP/fml_loader@4.0.6/net.neoforged.fml.loading.FMLLoader.beginModScan(FMLLoader.java:122)
	at MC-BOOTSTRAP/fml_loader@4.0.6/net.neoforged.fml.loading.FMLServiceProvider.beginScanning(FMLServiceProvider.java:84)
	at MC-BOOTSTRAP/cpw.mods.modlauncher@11.0.3/cpw.mods.modlauncher.TransformationServiceDecorator.runScan(TransformationServiceDecorator.java:97)
	at MC-BOOTSTRAP/cpw.mods.modlauncher@11.0.3/cpw.mods.modlauncher.TransformationServicesHandler.lambda$runScanningTransformationServices$7(TransformationServicesHandler.java:94)
```

Full Log: https://gist.github.com/robotgryphon/8aa3e40f801290a360e5d8dd2c352835

marchermans:
### TLDR:
- New FML hates multiple MCs per classpath
- Is not entirely preventable when multiple sourcesets are involved, specifically then when multiple MC providing plugins from different sourcesets are in play.
- Create handling logic for primary sourcesets.
- Fix tests
- Introduce support for config cache running with runs

### Changes:
- Introduce primary sourceset
  - Alters the behaviour of the classpath of a run. Does not change anything for IDEs, as those have their own behaviours when it comes to classpaths, only deals with run tasks.
- Remove RunExec, plain JavaExec now, which means support for runs in config cache
- Cleans up internal handling of config cache.
- Tests are fixed
- Removes the windows specific locking behaviour because it does not work with Daemons now that recompile is being cached.

### Issues:
Closes: #214 
Closes: #198


userdev/src/functionalTest/groovy/net/neoforged/gradle/userdev/CentralCacheTests.groovy
```diff
@@ -89,53 +91,6 @@ class CentralCacheTests extends BuilderBasedTestSpecification {
         cacheDir.listFiles().size() == 4
     }
 
-    def "cache_supports_running_gradle_in_parallel"() {
```
marchermans, line 92: I had to remove this test, because it is insanely flaky.
Not because of the central cache that it was testing that worked, but gradle itself simply not liking that it was running in parallel

common/src/main/java/net/neoforged/gradle/common/util/ClasspathUtils.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public static boolean isMinecraftClasspathEntry(File entry) {
+        if (!isClasspathEntry(entry)) {
+            return false;
+        }
+
+        //Check if the file contains the class:
+        //net.minecraft.client.Minecraft
+        //This is a class that is always present in the Minecraft jar.
```
sciwhiz12, line 27: Technically, this class is always present for the Minecraft _client_ JAR -- it won't be for the dedicated server JAR.

Would this cause problems in the future for a potential split-source system (where a source exists that's only the dedicated server)? Perhaps we should check for another prominent class that's likely to exist forevermore, such as `MinecraftServer` or `SharedConstants`?

marchermans: Sadly, that is the class FML checks itself for: https://github.com/neoforged/FancyModLoader/blob/89a32f970ba9137cb74e681af9cdaed0fc3e2085/loader/src/main/java/net/neoforged/fml/loading/targets/CommonUserdevLaunchHandler.java#L24

sciwhiz12: Ah. Then could you please add a comment linking to that, so we know in the future to keep the two in sync?

sciwhiz12 (dismissed):
Logic looks good to me. As long as the tests agree, automated and local/manual, should be good.

PR #221, opened 2024-06-28
[Fix]: Fixes IDEA not detecting preRun tasks

marchermans:
TLDR:
- This fixes an issue introduced in the last PR that cases IDEA not to see pre run tasks when running a run
- Supersedes: #220 because 220 re-introduces the bug that IDEA wants to run idePostSync after every task.

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/IdeManagementExtension.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/IdeManagementExtension.java
@@ -50,7 +50,7 @@ public abstract class IdeManagementExtension {
         // IntelliJ remembers to run the task post-sync even if the import fails. That will cause
         // situations where import errors (i.e. dependency resolution errors) will be masked by
         // the failed idePostSync task, since it was never created in that particular import.
-        if (isIdeaImport()) {
+        if (isIdeaAttached() && isIdeaSyncing()) {
             getOrCreateIdeImportTask();
         }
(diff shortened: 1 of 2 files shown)


ChampionAsh5357 (approved):

common/src/main/java/net/neoforged/gradle/common/extensions/IdeManagementExtension.java
```diff
@@ -62,7 +62,18 @@ public IdeManagementExtension(Project project) {
...
+        return Boolean.getBoolean("idea.active");
+    }
+
+    /**
+     * Get whether Gradle is being invoked through IntelliJ IDEA.
+     *
+     * <p>This can be through a project import, or a task execution.</p>
+     *
+     * @return whether this is an IntelliJ-based invocation
+     */
```
sciwhiz12, line 75: This javadoc is a duplicate of the one for `isIdeaAttached`. What is the difference between the two?

ChampionAsh5357: Synced is only set when the project is being loaded/reloaded within the IDE. Attached is always set to true when executing a task in an IDE.

marchermans: They are somewhat interchangeable yes. But they indicate different things.

PR #237, opened 2024-08-15
Do not version the asset directory

SquidDev:
Currently assets are written to `.gradle/caches/minecraft/assets/$MCVER`. However, the asset store is content-addressed, and so can be shared across versions. We drop the version suffix from the path to allow this.

I've deprecated the old two-arg `DownloadAssets.getAssetsDirectory` method in case other plugins/buildscripts are using it, but not sure if it counts as public API, so happy to just remove if preferred.


marchermans (approved):

SquidDev:
Thanks! :heart:

PR #249, opened 2024-10-20
[Feature/Fix]: Interface injections, and minor bugfixes

marchermans:
### TLDR:
- Implements interface injection
- Adds tests for interface injection
- **Removes** the ability for the gradle script to define ATs, as it never really worked and was a headache

---

### Changes
- Implement interface injection:
  - Implement a way to add additional source files
  - Remove the additional source file outputs from the compile result
  - Handle internal processing on the pipeline scope of additional processing files.
- Implement preprocessing of the recompile step
- Change the problem reporter to be logger based so that Gradle 8.10 does not fall on its face, there is now more then one kind of ProblemReporter implementation in the service stack, so it has no idea what to inject and I can't be bothered to figure that one out, so a simple logger with the same output will do for now.
  - This behaviour can be changed using a convention, and the legacy integrated reporter is still available if desired.


common/src/main/java/net/neoforged/gradle/common/interfaceinjection/InterfaceInjectionPublishing.java
```diff
@@ -0,0 +1,56 @@
...
+import org.gradle.api.attributes.AttributeContainer;
+import org.gradle.api.attributes.Category;
+import org.gradle.api.component.AdhocComponentWithVariants;
+
+public class InterfaceInjectionPublishing {
+
+    public static final String INTERFACE_INJECTION_ELEMENTS_CONFIGURATION = "InterfaceInjectionElements";
+    public static final String INTERFACE_INJECTION_API_CONFIGURATION = "InterfaceInjectionApi";
+    public static final String INTERFACE_INJECTION_CONFIGURATION = "InterfaceInjection";
+    public static final String INTERFACE_INJECTION_CATEGORY = "InterfaceInjection";
```
lukebemish: This should probably be lowercase, to match both MDG, which uses `interfaceinjection`, and gradle itself, for which all category attribute values are lowercase.

lukebemish (changes requested):

lukebemish:
Nope, still not it -- should be fully lowercase, `interfaceinjection`.

PR #255, opened 2024-11-18
[Fix]: Properly collect all dependencies in a configuration

marchermans:
When switching to detached configurations some configurations that are used as extension targets, extend other configurations themselves.

This was not taken into account and caused issues with the userdev artifact not containing all dependencies on NF.


marchermans:
Merging because tests are because the FML error message changed.

PR #256, opened 2024-12-12
Fix use of logger in ZIP inject source

shartte:
Fixes

```
Caused by: org.gradle.internal.service.UnknownServiceException: No service of type Logger available in project services.
    at org.gradle.internal.service.DefaultServiceRegistry.get(DefaultServiceRegistry.java:342)
```


marchermans (approved):

marchermans:
Overriding tests because it is missing the `data` test, which is now `clientData`

PR #257, opened 2024-12-14
Implement generation of split-source Manifest entries

shartte:
NFRT companion PR: https://github.com/neoforged/NeoFormRuntime/pull/50


shartte:
Alright, this also now supports the bundled server jar properly.

marchermans:
Overriding tests because it is missing the data test, which is now clientData

PR #261, opened 2025-01-18
[HotFix] Fix an issue with Gradle 8.10 and worker threads realising values on Properties

marchermans:
### TLDR:
When an IDEA import is ran, the runs are realised potentially on a different thread then the thread that owns a project, this causes a hard to debug error which is hidden in an IDEA warning.

By realising the properties eagerly, and capturing them in locals (Which is possible), the resulting error is not triggered.

### Further research:
- finalizeOnRead()? might help here, but it might also not as the runs where not realised on those properties before during an import.


marchermans:
Git F'ed itself with the tag checkout.
Merging as these test pass on different platforms.

PR #269, opened 2025-06-03
Update to gradle 13 and add config cache running.

marchermans:
## Changes:
- Bump to minimal gradle requirement to 13
- Add support for better handling of the library cache when running in caching mode.


shartte:
You really got me good Orion. "Update to Gradle 13"... I said out loud "WHAT? WE ARE AT GRADLE 13 ALREADY???" :-D

PR #316, opened 2026-05-21
Use `Project.ext` as a fallback for property lookups

isXander:
## What

I have made `WithPropertyLookup` source from a configuration-time snapshot value
of `Project.ext` as a fallback for property lookups, when `providers.gradleProperty()` has no value.

## Why

This change allows for programmatic configuration of NeoGradle conventions.

Such uses include: 

- Plugins that apply before NeoGradle may set conventions required for NeoGradle to correctly behave in conjunction with it.
- `build-logic` conventions plugins may set the same set of conventions across all applied projects.

### DIFF
--- a/common/src/main/java/net/neoforged/gradle/common/extensions/base/WithPropertyLookup.java
+++ b/common/src/main/java/net/neoforged/gradle/common/extensions/base/WithPropertyLookup.java
@@ -3,11 +3,13 @@ package net.neoforged.gradle.common.extensions.base;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 import org.gradle.api.file.Directory;
+import org.gradle.api.plugins.ExtraPropertiesExtension;
 import org.gradle.api.provider.Provider;
 
 import java.io.File;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 import static net.neoforged.gradle.dsl.common.util.Constants.SUBSYSTEM_PROPERTY_PREFIX;
 
@@ -19,7 +21,7 @@ public abstract class WithPropertyLookup {
     }
 
     protected Provider<String> getStringProperty(String propertyName, String defaultValue) {
-        final Provider<String> property = this.project.getProviders().gradleProperty(SUBSYSTEM_PROPERTY_PREFIX + propertyName);
+        final Provider<String> property = this.getProperty(SUBSYSTEM_PROPERTY_PREFIX + propertyName);
         if (defaultValue == null)
             return property;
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/neoforged/gradle/common/extensions/base/WithPropertyLookup.java
```diff
@@ -57,11 +59,29 @@ protected Provider<Boolean> getBooleanProperty(String propertyName) {
...
+        // Take a snapshot of the extra property at configuration as a fallback.
+        // Extra properties are intended to be provided from other Gradle plugins that are
+        // applied before this plugin to set default values programmatically.
+        ExtraPropertiesExtension ext = this.project.getExtensions().getExtraProperties();
+        String extValue;
+        if (ext.has(propertyName)) {
+            extValue = Objects.toString(ext.get(propertyName));
+        } else {
+            extValue = null;
+        }
```
marchermans, line 77: Would it not make sense to move this into the Supplier for the Provider?
That way it is only evaluated when we query it and not immediatly?

marchermans (changes requested):
I think we can improve this a little bit.

common/src/main/java/net/neoforged/gradle/common/extensions/base/WithPropertyLookup.java
isXander: This immediate evaluation is purposeful as I keep in mind configuration cache. I don't think it's safe to capture `Project` or `ExtraPropertiesContainer` within a provider like that.

Also, for the use cases where ext are used, they would already be available at configuration time.

marchermans: Okey good enough reason for me.

marchermans (approved):

lukebemish:
This suggestion breaks project isolation, and current Gradle work seems to be moving heavily away from project properties and especially from `project.ext` n general (I'll try and find the recent issues/PRs when I have a chance). I am heavily opposed to this suggestion.

Use of project properties in general is a bad idea. In addition to breaking project isolation in certain scenarios when loaded from subprojects (see: arch loom, for details), this whole scenario makes a bunch of logic now order-dependent that was entirely declarative before.

If the goal is simply to be able to configure them per-project -- you should prefer having a `Property<Boolean>` extension field somewhere that is `convention`-ed to the `gradleProperty`. My understanding is that most of these fields already are? If they need to be loaded too early for that -- a `Settings` extension to configure values on the project extensions would likely suffice.

